PR #39, opened 2022-08-16
Removed consts from component, using consts from armoapi-go

amirmalka:
Using consts defined in:
https://github.com/armosec/armoapi-go/blob/master/armotypes/helmconsts.go


dwertent (approved):
Did you update the BE to support the non-depreciated labels?

PR #61, opened 2023-01-25
add otel with uptrace client

matthyx:
Signed-off-by: Matthias Bertschy <matthias.bertschy@gmail.com>


matthyx:
@dwertent you can review this one, but don't merge before I confirm with the integration test

mainhandler/handlerequests.go
```diff
@@ -100,24 +104,27 @@ func (mainHandler *MainHandler) HandleRequest() []error {
...
 }
 
-func (mainHandler *MainHandler) HandleSingleRequest(sessionObj *utils.SessionObj) {
+func (mainHandler *MainHandler) HandleSingleRequest(ctx context.Context, sessionObj *utils.SessionObj) {
+	ctx, span := otel.Tracer("").Start(ctx, "mainHandler.HandleSingleRequest")
+	defer span.End()
 
 	status := "SUCCESS"
 
 	actionHandler := NewActionHandler(mainHandler.k8sAPI, sessionObj, mainHandler.commandResponseChannel)
```
dwertent, line 123: Add the `ctx` to the ActionHandler struct this way you can use it without the need of passing to the inner functions

dwertent (changes requested):

matthyx:
@dwertent done
I don't understand the GH action failure...

mainhandler/handlerequests.go
matthyx: I remember why that didn't work in kubescape and why I have decided to pass it around (despite the heaviness)... This is because if I want to create a sub span inside a func, I have to first save ctx and not forget to restore it before returning, otherwise spans would always be children of children (instead of sisters, aunts, ...).

matthyx:
@dwertent help, another strange CI error...

Daniel-GrunbergerCA (approved):

dwertent (approved):

PR #67, opened 2023-02-19
cleanUp flow

Daniel-GrunbergerCA:
## Overview

Add cleanUp routine

Missing integration with storage

https://github.com/kubescape/operator/pull/66

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -55,7 +55,7 @@ func (wh *WatchHandler) cleanUp(ctx context.Context) error {
 	wh.cleanUpIDs()
 
 	// build maps, retrieve current instanceIDs
-	currentInstanceIDs := wh.getInstanceIDsAndBuildMaps(ctx, podsList)
+	currentInstanceIDs := wh.listInstanceIDsAndBuildMaps(ctx, podsList)
 
 	// image IDs in storage that are not in current map should be deleted
 	imageIDsForDeletion := make([]string, 0)
@@ -81,7 +81,7 @@ func (wh *WatchHandler) cleanUp(ctx context.Context) error {
 
 // build internal maps from PodList, generate instanceIDs and return them.
 // Not using 'buildMaps' because we don't want to list Pods twice
-func (wh *WatchHandler) getInstanceIDsAndBuildMaps(ctx context.Context, podList *core1.PodList) []string {
+func (wh *WatchHandler) listInstanceIDsAndBuildMaps(ctx context.Context, podList *core1.PodList) []string {
 	instanceIDs := make([]string, 0)
 
 	for i := range podList.Items {
@@ -97,14 +97,14 @@ func (wh *WatchHandler) getInstanceIDsAndBuildMaps(ctx context.Context, podList 
 			continue
 		}
 
-		newInstanceIDs := wh.getInstanceIDsAndBuildMapsFromParentWlid(parentWlid, wl, &podList.Items[i])
+		newInstanceIDs := wh.listInstanceIDsAndBuildMapsFromParentWlid(parentWlid, wl, &podList.Items[i])
 		instanceIDs = append(instanceIDs, newInstanceIDs...)
 	}
 
 	return instanceIDs
 }
 
-func (wh *WatchHandler) getInstanceIDsAndBuildMapsFromParentWlid(wlid string, wl k8sinterface.IWorkload, pod *core1.Pod) []string {
+func (wh *WatchHandler) listInstanceIDsAndBuildMapsFromParentWlid(wlid string, wl k8sinterface.IWorkload, pod *core1.Pod) []string {
 	instanceIDs := make([]string, 0)
 	imgIDsToContainers := extractImageIDsToContainersFromPod(pod)
 	for imgID, containerName := range imgIDsToContainers {
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -39,18 +39,20 @@ type WatchHandler struct {
 func (wh *WatchHandler) cleanUp(ctx context.Context) {
 	storageImageIDs, err := wh.ListImageIDsFromStorage()
 	if err != nil {
+		// we should continue in order to clean up instanceIDs
 		logger.L().Ctx(ctx).Error("error to ListImageIDsFromStorage", helpers.Error(err))
 	}
 
 	storageInstanceIDs, err := wh.ListInstanceIDsFromStorage()
 	if err != nil {
+		// we should continue in order to clean up the internal maps
 		logger.L().Ctx(ctx).Error("error to ListInstanceIDsFromStorage", helpers.Error(err))
 	}
 
 	// list Pods, extract their imageIDs and instanceIDs
 	podsList, err := wh.k8sAPI.ListPods("", map[string]string{})
 	if err != nil {
-		logger.L().Ctx(ctx).Error("could not complet cleanUp routine: error to ListPods", helpers.Error(err))
+		logger.L().Ctx(ctx).Error("could not complete cleanUp routine: error to ListPods", helpers.Error(err))
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


watcher/watcher.go
```diff
@@ -20,7 +20,10 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 )
 
-const retryInterval = 3 * time.Second
+const (
+	retryInterval          = 3 * time.Second
+	cleanUpRoutineInterval = 10 * time.Minute
```
dwertent, line 25: [Required]: The cleanup interval should be configurable

watcher/watcher.go
```diff
@@ -31,6 +34,87 @@ type WatchHandler struct {
...
+	}
+
+	// list Pods, extract their imageIDs and instanceIDs
+	podsList, err := wh.k8sAPI.ListPods("", map[string]string{})
+	if err != nil {
+		return err
+	}
+
+	// reset maps
+	wh.cleanUpMaps()
```
dwertent, line 56: [Question]: Why did you call the function `cleanUpMaps` and then add a comment saying "reset maps", why not name the function `resetMaps` or something similar?
Personally, I think such a function should not be called based on the data structure. Meaning, the data structure is of a type map (dictionary) that is representing the unique imagesIDs and instancesIDs, so the function can be called `resetIDs` or even `resetIDsStructures`. Think about it...

watcher/watcher.go
```diff
@@ -31,6 +34,87 @@ type WatchHandler struct {
...
+	currentInstanceIDs := wh.getInstanceIDsAndBuildMaps(ctx, podsList)
```
dwertent, line 57: [TODO]: If this function returns a map/s, there us now reason to add the "builds maps" in it's name.
If the get function builds other maps that are not related directly to the "get" operation, you should consider removing the "get" from the function name or splitting this to "building maps" and "get".

watcher/watcher.go
```diff
@@ -31,6 +34,87 @@ type WatchHandler struct {
...
+
+	// image IDs in storage that are not in current map should be deleted
+	imageIDsForDeletion := make([]string, 0)
+	for _, imageID := range storageImageIDs {
+		if _, ok := wh.imagesIDToWlidsMap[imageID]; !ok {
```
dwertent, line 62: [Question]: You don't need to synchronize read/write access to the `wh.imagesIDToWlidsMap` object?

watcher/watcher.go
dwertent, line 57: [Required]: Also, it seams like the function returns a `list` object, if so, you should name it `list` and not `get`.

watcher/watcher.go
```diff
@@ -31,6 +34,87 @@ type WatchHandler struct {
...
+			instanceIDsForDeletion = append(instanceIDsForDeletion, instanceID)
+		}
+	}
+
+	return wh.deleteInstanceIDsFromStorage(instanceIDsForDeletion)
+}
+
+// build internal maps from PodList, generate instanceIDs and return them.
+// Not using 'buildMaps' because we don't want to list Pods twice
+func (wh *WatchHandler) getInstanceIDsAndBuildMaps(ctx context.Context, podList *core1.PodList) []string {
```
dwertent, line 83:
```suggestion
func (wh *WatchHandler) listInstanceIDsAndBuildMaps(ctx context.Context, podList *core1.PodList) []string {
```
I commented earlier about this function name and about splitting it into two.

watcher/watcher.go
```diff
@@ -51,9 +135,40 @@ func NewWatchHandler(ctx context.Context) (*WatchHandler, error) {
...
 
 	wh.currentPodListResourceVersion = podsList.GetResourceVersion()
+
+	wh.startCleanerRoutine(ctx)
+
 	return wh, nil
 }
 
+// start routine which cleans up unused imageIDs and instanceIDs from storage, and triggers relevancy scan
+func (wh *WatchHandler) startCleanerRoutine(ctx context.Context) {
```
dwertent, line 145: [TODO]: Consider renaming this function because you are calling here the `triggerRelevancyScan` function.

watcher/watcher.go
```diff
@@ -88,6 +203,30 @@ func (wh *WatchHandler) ListImageIDsFromStorage() ([]string, error) {
...
+	defer wh.imageIDsMapMutex.Unlock()
+
+	wh.imagesIDToWlidsMap = make(map[string][]string)
+}
+
+func (wh *WatchHandler) cleanUpWlidsToContainerToImageIDMap() {
+	wh.wlidsToContainerToImageIDMapMutex.Lock()
+	defer wh.wlidsToContainerToImageIDMapMutex.Unlock()
+
+	wh.wlidsToContainerToImageIDMap = make(map[string]map[string]string)
```
dwertent, line 222: [TODO]: `map[string]map[string]string` **consider** defining a struct instead of having a map within a map.

dwertent (changes requested):

watcher/watcher.go
Daniel-GrunbergerCA: Regarding your question, the `cleanUpMaps` does what its name states - cleans up the maps. The comment is related to the flow, meaning we clean the maps then build them again, that's why I wrote reset maps. I will try to make it more clear in the comments. 
I will also rename the function

watcher/watcher.go
Daniel-GrunbergerCA: yes

watcher/watcher.go
```diff
@@ -20,26 +20,82 @@ import (
...
+	instanceIDsMutex                  *sync.RWMutex
+	wlidsToContainerToImageIDMapMutex *sync.RWMutex
 	currentPodListResourceVersion     string // current PodList version, used by watcher (https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes)
 }
 
+// remove unused imageIDs and instanceIDs from storage. Update internal maps
+func (wh *WatchHandler) cleanUp(ctx context.Context) {
+	storageImageIDs, err := wh.ListImageIDsFromStorage()
+	if err != nil {
+		logger.L().Ctx(ctx).Error("error to ListImageIDsFromStorage", helpers.Error(err))
```
matthyx, line 42: does it make sense to continue after this error?

watcher/watcher.go
```diff
@@ -20,26 +20,82 @@ import (
...
+	}
+
+	storageInstanceIDs, err := wh.ListInstanceIDsFromStorage()
+	if err != nil {
+		logger.L().Ctx(ctx).Error("error to ListInstanceIDsFromStorage", helpers.Error(err))
```
matthyx, line 47: here too

watcher/watcher.go
Daniel-GrunbergerCA: IMO it makes sense, since on the continuation we will list all Pods and according to that build the internal map. That way workloads that have been deleted will be erased from the map. 
If I get an error there (listing Pods), then I won't continue

matthyx: pls add a comment with justification

watcher/watcher_test.go
```diff
@@ -16,292 +15,293 @@ func NewWatchHandlerMock() *WatchHandler {
...
-							},
-						},
-					}}},
-			expectedImageIDsMap: map[string][]string{
-				"alpine@sha256:1": {pkgwlid.GetWLID("", "default", "pod", "test")},
-				"alpine@sha256:2": {pkgwlid.GetWLID("", "default", "pod", "test")},
-			},
-		},
-	}
+// func TestBuildImageIDsToWlidsMap(t *testing.T) {
```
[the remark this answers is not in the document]
Daniel-GrunbergerCA: So according to @dwertent review I added an internal map of `instanceIDs`. This caused me to use K8S API inside the buildMap function, which makes me enable to unit test it. I put it on comment since there is a chance this map will be removed, and then I can return to the old function and its unit tests

watcher/watcher.go
matthyx: pls add a comment with justification

watcher/watcher_test.go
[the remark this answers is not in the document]
matthyx: ok good

matthyx (approved):

dwertent (approved):

PR #74, opened 2023-03-08
use RegistryProvider when set to choose registry kind

matthyx:
## Overview
Fixes SUB-689 when a user tries to add registry scanning for a Harbor registry.
This probably failed for other registries where the URL didn't allow proper type detection.

### DIFF
--- a/mainhandler/imageregistryhandler.go
+++ b/mainhandler/imageregistryhandler.go
@@ -607,7 +607,6 @@ func (registryScan *registryScan) setRegistryKind() {
 
 func (registryScan *registryScan) getRegistryProvider() string {
 	if strings.Contains(registryScan.registryInfo.RegistryName, ".dkr.ecr") {
-
 		return "ecr"
 	}
 	if strings.Contains(registryScan.registryInfo.RegistryName, "gcr.io") {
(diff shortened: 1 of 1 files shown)


mainhandler/imageregistryhandler.go
```diff
@@ -606,8 +606,10 @@ func (registryScan *registryScan) setRegistryKind() {
 }
 
 func (registryScan *registryScan) getRegistryProvider() string {
+	if registryScan.registryInfo.RegistryProvider != "" {
+		return registryScan.registryInfo.RegistryProvider
```
amirmalka, line 610: the only thing that I'm concerned about when this is the first `if` statement is whether it will impact the other conditionals... maybe it should be the last one? before returning empty

amirmalka (approved):

mainhandler/imageregistryhandler.go
```diff
@@ -616,7 +615,7 @@ func (registryScan *registryScan) getRegistryProvider() string {
 	if strings.Contains(registryScan.registryInfo.RegistryName, "quay.io") {
 		return "quay.io"
 	}
-	return ""
+	return registryScan.registryInfo.RegistryProvider
```
matthyx, line 618: ok @amirmalka let's do something like this

PR #81, opened 2023-03-17
Merge master to relevancy

dwertent:
## Overview
This way we can test relevancy with the latest updates


main.go
```diff
@@ -35,11 +35,12 @@ func main() {
 		return
 	}
 
-	// to enable otel, set OTEL_COLLECTOR_SVC=otel-collector:4317
+	// to enable otel, set OTEL_COLLECTOR_SVC=otel-collector:4317
```
vladklokun, line 38: style: there’s an extra whitespace

vladklokun (approved):
LGTM

main.go
vladklokun: As you mentioned, we will soon have a linter that will catch these. Never mind this.

PR #82, opened 2023-03-18
feat: schedule CVE scans for new SBOMs

vladklokun:
## Overview

This PR adds support for scheduling CVE scans for newly detected SBOMs, as required by the relevancy feature.

## How to Test

To test, install the `relevancy` Kubescape Helm chart with the `kubescapeStorage.enabled=true` and observe the behavior of the Operator.

put an [x] in the box to get it checked 


watcher/watcher_test.go
```diff
@@ -21,6 +29,443 @@ func NewWatchHandlerMock() *WatchHandler {
...
+			assert.Equalf(t, tc.expectedErrors, actualErrors, "Errors should match")
+		})
+	}
+}
+
+func TestSBOMWatch(t *testing.T) {
+	t.Skipf(
+		"vladklokun: blocks and deadlocks while listening on the sbomWatcher.ResultChan(). " +
+		"Does not reproduce in a live cluster on a live Watch() object",
+	)
```
vladklokun, line 422: thought: I thought this note might be useful for posterity

So that we remember that there is a test that is not passing because the test deadlocks, and if we ever get a chance, we should refactor the function to make it more testable.

I think we can fix this test if we create a `NewWatcherFunc() watch.Interface` test double, include a bi-directional channel that we own there, and then close it after the test.

Or we can employ a "shut down immediately" channel for the `SBOMWatch` function, or do some tricks with `context.Context`.

watcher/watcher.go
```diff
@@ -179,9 +189,104 @@ func (wh *WatchHandler) getImagesIDsToWlidMap() map[string][]string {
...
+	var watcher watch.Interface
+	var err error
+	for {
+		select {
+		case sbomEvent, ok := <-sbomEvents:
+			if ok {
+				inputEvents <- sbomEvent
+			} else {
+				sbomWatcherUnavailable <- struct{}{}
+			}
```
vladklokun, line 257: thought: this repeats multiple times because we both have to handle `sbomEvents` and `errorCh` closing, and also handle the watcher being unavailable

As an alternative, we might use one more channel which clearly signals that the input `sbomEvents` channel has closed and rely on the fact that we assume it cannot be close by anything other than the watcher becoming unavailable.

But then, testing `SBOMWatch` proves difficult, since it always tries to reconcile a closed Watch.

watcher/watcher.go
```diff
@@ -179,9 +189,104 @@ func (wh *WatchHandler) getImagesIDsToWlidMap() map[string][]string {
...
+		// existing ones or other operations
+		if event.Type != watch.Added {
+			continue
+		}
+
+		imageID := obj.ObjectMeta.Name
+
+		wh.imageIDsMapMutex.RLock()
+		wlids, ok := wh.imagesIDToWlidsMap[imageID]
+		wh.imageIDsMapMutex.RUnlock()
```
vladklokun, line 217: thought: we should really abstract the image IDs to WLIDs map into a separate struct

This was the reason behind many nasty deadlocks I have encountered, especially with how often we like to use `defer mu.RUnlock()`.

Another note here is that, in my opinion, Pod, SBOM and SBOM Filtered watches should be separate structs that share the image ID to WLIDs map or some other map they have. Because so far it seems that the `WatchHandler` struct does too many things.

watcher/watcher.go
```diff
@@ -14,18 +15,26 @@ import (
...
 	"golang.org/x/exp/slices"
 	core1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/watch"
 )
 
+var (
+	ErrUnsupportedObject = errors.New("Unsupported object type")
+	ErrUnknownImageID    = errors.New("Unknown image ID")
+)
```
vladklokun, line 29: thought: we should ideally make these custom error types, so we can store information inside them

Like image IDs, the type of the unsupported object and maybe its payload.

mainhandler/handlerequests.go
```diff
@@ -85,7 +87,15 @@ func (mainHandler *MainHandler) HandleWatchers(ctx context.Context) {
...
 
-	watchHandler, err := watcher.NewWatchHandler(ctx)
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		panic(err)
+	}
+	ksStorageClient, err := kssc.NewForConfig(config)
+	if err != nil {
+		panic(err)
+	}
```
vladklokun, line 97: question: can we safely assume that the operator is running inside a cluster?

Or do we have to use some `k8sinterface` function to load the config?

And if we can’t load a config, it reasonable to immediately abort here? Since without the config, running the Operator seems kinda pointless.

vladklokun (commented):
Left some comments on possible future improvements and things where I cut corners. Please, feel free to address them.

mainhandler/handlerequests.go
dwertent: 1. We should not assume we are running in a cluster just for the purpose of debugging locally.
2. Yes, if the config is not found at all we should panic. In such a case we should call the `logger.L().Fatal()` function, not `panic`.

mainhandler/handlerequests.go
```diff
@@ -85,7 +87,15 @@ func (mainHandler *MainHandler) HandleWatchers(ctx context.Context) {
 		}
 	}()
 
-	watchHandler, err := watcher.NewWatchHandler(ctx)
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		panic(err)
+	}
+	ksStorageClient, err := kssc.NewForConfig(config)
```
dwertent, line 94: Here you can call the `k8sinterface.GetK8sConfig()` function, this is what I expect to see here:
```suggestion
	ksStorageClient, err := kssc.NewForConfig(k8sinterface.GetK8sConfig())
```
The `rest.InClusterConfig()` can be removed

dwertent, line 94: I will consider adding the storage client initialization to the `NewKubernetesApi()` function.

dwertent (changes requested):

mainhandler/handlerequests.go
vladklokun: Addressed in the latest commit.

vladklokun: Why?

It seems to me that the Storage client is pretty much stand-alone, not tied to the `KubernetesApi` interface, and very testable on its own. I am concerned that tying it to the KubernetesApi interface would make it harder to use Storage in tests while providing debatable additional benefits.

dwertent (approved):

PR #85, opened 2023-03-22
change api and fix readiness

Daniel-GrunbergerCA:
## Overview

1 - Use only "scan" API (also for SBOM calculation)
2 - Fix call to vuln-scan readiness probe

### DIFF
--- a/mainhandler/handlerequestsutils.go
+++ b/mainhandler/handlerequestsutils.go
@@ -2,7 +2,6 @@ package mainhandler
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/kubescape/go-logger"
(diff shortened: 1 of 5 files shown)


mainhandler/handlerequestsutils.go
```diff
@@ -65,8 +64,6 @@ func isActionNeedToWait(action apis.Command) waitFunc {
 
 func waitForVulnScanReady() {
 	fullURL := getVulnScanURL()
-	// replace port
-	fullURL.Host = strings.ReplaceAll(fullURL.Host, fullURL.Port(), probes.ReadinessPort)
```
Daniel-GrunbergerCA, line 69: We were using a hardcoded port which was on another package..

matthyx (approved):

PR #87, opened 2023-03-22
fixes

Daniel-GrunbergerCA:
## Overview

1 - Add `imageID` to log
2 - Lookup parent workload up untill Node
3 - Remove duplicate log ("vuln scan is ready")

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -530,7 +530,12 @@ func (wh *WatchHandler) getParentWorkloadForPod(pod *core1.Pod) (workloadinterfa
 	if err != nil {
 		return nil, err
 	}
+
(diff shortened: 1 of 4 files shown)


watcher/watcher.go
```diff
@@ -530,7 +530,12 @@ func (wh *WatchHandler) getParentWorkloadForPod(pod *core1.Pod) (workloadinterfa
 	if err != nil {
 		return nil, err
 	}
+
 	kind, name, err := wh.k8sAPI.CalculateWorkloadParentRecursive(wl)
+	if kind == "Node" {
```
Daniel-GrunbergerCA, line 535: I have not fixed changed `CalculateWorkloadParentRecursive` since the Node is indeed the parent, and the function does what is supposed to do. We just have a specific use-case of vulnerability scanning, where we don't consider the Node a parent.

PR #93, opened 2023-03-24
refactor: encapsulate ImageIDs to WLIDs map 

vladklokun:
## Overview
This PR encapsulates a map-and-mutex map of Image hashes to WLIDs in a thread-safe map-like type, so that they can be accessed more conveniently, with less leeway to break synchronization.

## How to Test

To check unit tests, run:
```
go test -v ./watcher -race -failfast -count=1000
```

Other than that, run whatever Operator flows you deem necessary. 


dwertent (approved):
Well done. I liked the refactoring you did with the lock and unlock :)

PR #99, opened 2023-03-26
Update instanceID

dwertent:
## Overview
We are using instancesID (podSpecID) as defined [here](https://github.com/kubescape/k8s-interface/blob/update-instanceid/instanceidhandler/interface.go)


go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/websocket v1.5.0
 	github.com/kubescape/go-logger v0.0.11
-	github.com/kubescape/k8s-interface v0.0.112
+	github.com/kubescape/k8s-interface v0.0.114-0.20230326195717-d11a8f35efc2
```
matthyx, line 20: you don't want to wait for the released version?

mainhandler/vulnscan.go
```diff
@@ -287,6 +287,15 @@ func (actionHandler *ActionHandler) scanWorkload(ctx context.Context, sessionObj
 		logger.L().Info(fmt.Sprintf("workload %s has no podSpec, skipping", actionHandler.wlid))
 		return nil
 	}
+
+	// get pod instanceID
+	instancesID, err := instanceidhandler.GenerateInstanceIDFromPod(pod)
```
matthyx, line 292: instanceIDs ? I'm not sure about plurals for composed words.

matthyx (commented):
maybe let's wait for the released lib

mainhandler/vulnscan.go
dwertent: Resolved in [this](https://github.com/kubescape/operator/pull/99/commits/7f88b33efd8b0c5d9c0d7d00961022b607d96311) commit

matthyx (approved):

PR #115, opened 2023-03-28
fix: do not block on sending watcher down notifications

vladklokun:
## Overview
Previously, WatchHandler tried to send a notification that the watcher was down in the same thread of execution. Sending to a channel blocked indefinitely.

This change sends in a separate goroutine, so sending to a channel would not block and can be processed with `select`.


matthyx (approved):
much nicer this way

Daniel-GrunbergerCA (approved):

PR #120, opened 2023-03-28
feat: delete unrecognized SBOM filtereds

vladklokun:
## Overview

Watch for Filtered SBOMs and delete those we do not recognize.

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -25,14 +25,18 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 )
 
+const (
+	retryInterval = 3 * time.Second
+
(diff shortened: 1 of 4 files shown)


watcher/watcher.go
```diff
@@ -172,10 +176,43 @@ func (wh *WatchHandler) getWlidsToContainerToImageIDMap() WlidsToContainerToImag
 	return wh.wlidsToContainerToImageIDMap
 }
 
+func labelsToInstanceID(labels map[string]string) (string, error) {
```
dwertent, line 179: Let's call the function `extractInstanceID`

dwertent (approved):

PR #131, opened 2023-05-04
dont scan cronjobs && fix nil pointer

Daniel-GrunbergerCA:
## Overview

Operator was crashing because of a nil pointer access. This was done when logging properties of the instanceID, which was nil. 
After a deeper look, the conclusion is that the instanceID being generated is incorrect, due to the fact that we cannot find Pods which belong to a CronJob in a top-down approach. 
This PR:
1 - Fixes the nil pointer access
2 - Don't send command to scan CronJobs when generating commands for a namespace scan
3 - In case a command was given to scan a CronJob, it will be skipped

### DIFF
--- a/mainhandler/getworkload.go
+++ b/mainhandler/getworkload.go
@@ -27,15 +27,17 @@ func listWorkloadImages(workload k8sinterface.IWorkload, instanceIDs []instancei
 		id := getContainerID(instanceIDs, containers[i].Name)
 		if id != nil {
 			c = id.GetHashed()
+			containersData = append(containersData,
+				ContainerData{
+					image:     containers[i].Image,
+					container: containers[i].Name,
+					id:        c,
+				},
+			)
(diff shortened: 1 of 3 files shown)


mainhandler/getworkload.go
```diff
@@ -27,15 +27,17 @@ func listWorkloadImages(workload k8sinterface.IWorkload, instanceIDs []instancei
...
 		if id != nil {
 			c = id.GetHashed()
+			containersData = append(containersData,
+				ContainerData{
+					image:     containers[i].Image,
+					container: containers[i].Name,
+					id:        c,
+				},
+			)
+			logger.L().Debug("instanceID", helpers.String("str", id.GetStringFormatted()), helpers.String("id", id.GetHashed()), helpers.String("workloadID", workload.GetID()), helpers.String("container", containers[i].Name), helpers.String("image", containers[i].Image))
```
Daniel-GrunbergerCA, line 37: panic was here (id was nil)

matthyx (approved):

PR #137, opened 2023-06-06
fix(watcher): look for SBOMFiltered instanceIDs in the annotations

vladklokun:
## Overview
Previously the watcher assumed that the instance ID of a FitleredSBOM would be stored in the resource name. Now, as the naming scheme changed, the watcher relies on having it in the annotations.

## Additional Information
Ideally, fetching the instance ID should be available from either our custom types via a dedicated method, e.g `SBOMSPDXv2p3Filtered.InstanceID()` or extracted to the `k8s-interface` package. But doing this is faster.

### DIFF
--- a/watcher/watcher.go
+++ b/watcher/watcher.go
@@ -27,8 +27,6 @@ import (
 
 const (
 	retryInterval = 3 * time.Second
-
-	instanceIDAnnotationKey = "instanceID"
 )
 
 var (
@@ -119,7 +117,7 @@ func (wh *WatchHandler) GetWlidsToContainerToImageIDMap() WlidsToContainerToImag
 }
 
 func labelsToInstanceID(labels map[string]string) (string, error) {
-	instanceID, ok := labels[instanceIDAnnotationKey]
+	instanceID, ok := labels[instanceidhandlerv1.InstanceIDMetadataKey]
 	if !ok {
 		return instanceID, ErrMissingInstanceIDAnnotation
 	}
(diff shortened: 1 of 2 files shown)


vladklokun:
> It would be more reliable to check the annotation with the same name...

That’s exactly what I’m doing here. I understand the confusion came from the `labels` parameter of the function. I will rename this.

vladklokun:
I also want to add validation to the Storage APIServer that will ensure that all the values we expect in the annotations / labels are there. So you will help me a lot by providing a list of these values.

Can I expect that all key names [defined](https://github.com/kubescape/k8s-interface/blob/d5cd1f8a3b4b076c2535bb66f9cd0a86beab1bc7/instanceidhandler/v1/instanceidhandler.go#L13) in the k8s interface should be included when creating SBOMs, Filtered SBOMs and Vulnerability Manifests?

watcher/watcher.go
```diff
@@ -118,8 +116,8 @@ func (wh *WatchHandler) GetWlidsToContainerToImageIDMap() WlidsToContainerToImag
 	return wh.wlidsToContainerToImageIDMap
 }
 
-func labelsToInstanceID(labels map[string]string) (string, error) {
-	instanceID, ok := labels[instanceIDAnnotationKey]
+func annotationsToInstanceID(annotations map[string]string) (string, error) {
```
Daniel-GrunbergerCA, line 119: can't we make this function more generic (for every key we expect to be on annotations)? Or this is just temporary until we have an implementation in another pkg?

vladklokun: It’s a combination of both. Ideally, we want to get instance IDs by calling a method on the object we receive, like ` SBOMSPDXv2p3.InstanceID()`. But this might take more effort than we are prepare to spend, like re-generating the APIServer code, updating all the dependents etc. Instead, we are adding this small function to the Operator for now.

The next thing is that the Operator needs to get only the instance ID. It doesn’t care for other keys, so why would we spend extra time to write tests and implementation for other keys? It’s not like we are exporting this method and someone else is going to use it. So I think this is enough for us.

matthyx (approved):

PR #146, opened 2023-06-28
support in "security" fw

kooomix:


dwertent (changes requested):
"append" is the action the Operator will take in case you wish to scan the security framework.
I will name the environment variable and the representative object something like "trigger security framework".

kooomix:
> "append" is the action the Operator will take in case you wish to scan the security framework. I will name the environment variable and the representative object something like "trigger security framework".

@dwertent 
Done

dwertent (approved):

PR #154, opened 2023-08-08
Replace channel with concurrency

dwertent:
## Overview
Replaces channel with concurrency.
Set the desired number with the `WORKER_CONCURRENCY` environment variable.

### DIFF
--- a/utils/globalvar_test.go
+++ b/utils/globalvar_test.go
@@ -1,0 +1,62 @@
+package utils
+
+import (
+	"os"
+	"testing"
+)
+
+func TestParseIntEnvVar(t *testing.T) {
+	tests := []struct {
+		name         string
+		varName      string
+		varValue     string
+		defaultValue int
+		expected     int
+		expectError  bool
+	}{
+		{
+			name:         "environment variable not set",
+			varName:      "TEST_ENV_VAR_NOT_SET",
+			defaultValue: 10,
+			expected:     10,
+			expectError:  false,
+		},
+		{
+			name:         "environment variable set with valid integer",
+			varName:      "TEST_ENV_VAR_VALID",
+			varValue:     "20",
+			defaultValue: 10,
+			expected:     20,
+			expectError:  false,
+		},
+		{
+			name:         "environment variable set with invalid integer",
+			varName:      "TEST_ENV_VAR_INVALID",
+			varValue:     "invalid",
+			defaultValue: 10,
+			expected:     10,
+			expectError:  true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.varValue != "" {
+				os.Setenv(tt.varName, tt.varValue)
+			} else {
+				os.Unsetenv(tt.varName)
+			}
+
+			val, err := parseIntEnvVar(tt.varName, tt.defaultValue)
+			if tt.expectError && err == nil {
+				t.Fatalf("Expected an error, got nil")
+			} else if !tt.expectError && err != nil {
+				t.Fatalf("Expected no error, got %v", err)
+			}
+
+			if val != tt.expected {
+				t.Fatalf("Expected value %d, got %d", tt.expected, val)
+			}
+		})
+	}
+}
--- a/utils/globalvar_test.go
+++ b/utils/globalvar_test.go
@@ -1,7 +1,6 @@
 package utils
 
 import (
-	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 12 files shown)


utils/types.go
```diff
@@ -18,3 +20,24 @@ type CredStruct struct {
...
+type Job struct {
+	ctx        context.Context
+	sessionObj SessionObj
+}
+
+func (j *Job) GetContext() context.Context {
+	return j.ctx
+}
+
+func (j *Job) GetObj() SessionObj {
```
vladklokun, line 33: nitpick: getters in Go [should not](https://go.dev/doc/effective_go#Getters) be named `Get`

>  If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner.

```suggestion
func (j *Job) Context() context.Context {
	return j.ctx
}

func (j *Job) Obj() SessionObj {
```

vladklokun (commented):
Other than the comments I’ve left, LGTM.

utils/globalvars.go
```diff
@@ -59,6 +60,21 @@ func LoadEnvironmentVariables(ctx context.Context) (err error) {
 			CleanUpRoutineInterval = dur
 		}
 	}
+	ConcurrencyWorkers, _ = ParseIntEnvVar(ConcurrencyEnvironmentVariable, ConcurrencyWorkers)
 
 	return nil
 }
+
+func ParseIntEnvVar(varName string, defaultValue int) (int, error) {
```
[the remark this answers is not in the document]
dwertent: This function was [copied as-is from the KS repo](https://github.com/kubescape/kubescape/blob/cf08daf7fbb17c799325927f9f0898df05aa4b22/core/cautils/strutils.go#L53).
Since we don't have a global pkg for such functions, I simply inserted it here.
I will change the function to private and add unit tests.

utils/globalvar_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+			val, err := parseIntEnvVar(tt.varName, tt.defaultValue)
+			if tt.expectError && err == nil {
+				t.Fatalf("Expected an error, got nil")
+			} else if !tt.expectError && err != nil {
+				t.Fatalf("Expected no error, got %v", err)
+			}
+
+			if val != tt.expected {
+				t.Fatalf("Expected value %d, got %d", tt.expected, val)
```
vladklokun, line 58: issue: test case does not use the conventional `testify/assert`.

Did ChatGPT or something of the sort write this? We commonly use `testify/assert`, its assertions are simpler than that, so there’s no need for ifs and `t.Fatalf`.


```go
tests := []struct{
  name string
  inputVarName string
  inputVarValue string
  defaultValue int
  expectedValue int
  expectedError error
}{
  for _, tc := range tests {
    if tc.varValue != "" {
      t.Setenv(tt.varName, tt.varValue)
    } else {
      t.Setenv(tt.varName, "")
    }

    val, err := parseIntEnvVar(tt.varName, tt.defaultValue)

    assert.Equalf(t, expectedError, gotError)
    assert.Equalf(t, expectedValue, gotValue)
  }
}
```

utils/globalvar_test.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.varValue != "" {
+				os.Setenv(tt.varName, tt.varValue)
+			} else {
+				os.Unsetenv(tt.varName)
+			}
```
vladklokun, line 48: issue: an environment variable leaks after running this test


Since there is no cleanup of variables after setting them up, an environment variable will leak into the runner system. Which is why it’s a good idea to use test doubles for environment variables, not actual the actual live ones. If we do want the live variables, just use the function provided for testing:
```suggestion
			if tt.varValue != "" {
				t.Setenv(tt.varName, tt.varValue)
			} else {
				t.Setenv(tt.varName, "")
			}
```

vladklokun (changes requested):
Other than the test, LGTM.

utils/globalvar_test.go
dwertent: I implemented it with defer

utils/globalvar_test.go
```diff
@@ -0,0 +1,53 @@
...
+			defaultValue: 10,
+			expected:     10,
+			expectError:  true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv(tt.varName, tt.varValue)
+			defer t.Setenv(tt.varName, "")
```
vladklokun, line 46: nitpick(non-blocking): there’s no need to use defer `testing.T.Setenv` rolls back your change at the end of test automatically

vladklokun (approved):
LGTM!

PR #155, opened 2023-08-08
use distroless base image

matthyx:
this one requires a modification in the Helm chart:
```
securityContext:
  fsGroup: 65532
  runAsUser: 65532
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,8 +1,6 @@
-FROM --platform=$BUILDPLATFORM golang:1.20-alpine as builder
+FROM --platform=$BUILDPLATFORM golang:1.20-bullseye as builder
 
-ENV GO111MODULE=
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -11,11 +9,10 @@ RUN --mount=target=. \
     --mount=type=cache,target=/go/pkg \
     GOOS=$TARGETOS GOARCH=$TARGETARCH go build -o /out/operator .
 
-FROM alpine
+FROM gcr.io/distroless/static-debian11:nonroot
 
-RUN addgroup -S ks && adduser -S ks -G ks
-USER ks
-WORKDIR /home/ks/
+USER nonroot
```
dwertent, line 14: @matthyx you should make sure the operator will load the config from the correct path (in the helm chart)

dwertent (approved):

PR #168, opened 2023-09-04
fix running without serviceDiscovery

matthyx:
## PR Type:
Enhancement

___
## PR Description:
This pull request enhances the service discovery mechanism and error reporting in the application. It introduces a check for the 'ServiceDiscovery' component's enabled status before attempting to discover URLs. If the 'ServiceDiscovery' component is not enabled, the application will not attempt to discover URLs, preventing potential errors. Additionally, the error reporting mechanism has been updated to check if the 'eventReceiverRestURL' is not empty before sending error reports. This ensures that error reports are only sent when there is a valid URL to send them to. The backend version has also been bumped for a goroutine fix.

___
## PR Main Files Walkthrough:
`config/config.go`: Added 'ServiceDiscovery' to the 'Components' struct.
`main.go`: Added a check for 'ServiceDiscovery' component's enabled status before attempting to discover URLs.
`mainhandler/handlerequests.go`: Updated error reporting to check if 'eventReceiverRestURL' is not empty before sending error reports.
`mainhandler/imageregistryhandler.go`: Updated warning reporting to check if 'eventReceiverRestURL' is not empty before sending warning reports.
`mainhandler/kubescapehandler.go`: Updated details reporting to check if 'eventReceiverRestURL' is not empty before sending details reports.
`mainhandler/vulnscan.go`: Updated status and details reporting to check if 'eventReceiverRestURL' is not empty before sending status and details reports.
`restapihandler/triggeraction.go`: Updated error reporting to check if 'eventReceiverRestURL' is not empty before sending error reports.
`README.md`: Removed 'eventReceiverRestURL', 'eventReceiverWebsocketURL', and 'rootGatewayURL' from the example configuration.
`go.mod`: Bumped the version of 'github.com/kubescape/backend' from 'v0.0.1' to 'v0.0.4'.


README.md
```diff
@@ -30,9 +30,6 @@ If you running the Operator as part of the Kubescape cluster components, you nee
 	       "gatewayRestURL": "127.0.0.1:8002",
 	       "kubevulnURL": "127.0.0.1:8081",
 	       "kubescapeURL": "127.0.0.1:8080",
-	       "eventReceiverRestURL": "https://report.armo.cloud",
```
dwertent, line 33: You can add here the `serviceDiscovery`

mainhandler/handlerequests.go
```diff
@@ -233,15 +233,15 @@ func (mainHandler *MainHandler) HandleSingleRequest(ctx context.Context, session
 
 	actionHandler := NewActionHandler(mainHandler.clusterConfig, mainHandler.cfg, mainHandler.components, mainHandler.k8sAPI, sessionObj, mainHandler.commandResponseChannel, mainHandler.eventReceiverRestURL)
 	actionHandler.reporter.SetActionName(string(sessionObj.Command.CommandName))
-	actionHandler.reporter.SendDetails("Handling single request", true, sessionObj.ErrChan)
+	actionHandler.reporter.SendDetails("Handling single request", mainHandler.eventReceiverRestURL != "", sessionObj.ErrChan)
```
dwertent, line 236: Don't kill me, but I think the right approach here is to have a mock reporter so you initialize it when the server URL is not set.
What do you think?

main.go
```diff
@@ -43,13 +43,16 @@ func main() {
 	}
 	logger.L().Debug("loaded config for components", helpers.Interface("components", components))
 
-	services, err := config.GetServiceURLs("/etc/config/services.json")
-	if err != nil {
-		logger.L().Ctx(ctx).Fatal("failed discovering urls", helpers.Error(err))
-	}
+	var eventReceiverRestURL string
+	if components.Components.ServiceDiscovery.Enabled {
+		services, err := config.GetServiceURLs("/etc/config/services.json")
```
dwertent, line 48: We should support loading the config path from an env as well (not blocking)

main.go
```diff
@@ -43,13 +43,16 @@ func main() {
...
+			logger.L().Ctx(ctx).Fatal("failed discovering urls", helpers.Error(err))
+		}
 
-	eventReceiverRestURL := services.GetReportReceiverHttpUrl()
-	logger.L().Debug("setting eventReceiverRestURL", helpers.String("url", eventReceiverRestURL))
+		eventReceiverRestURL := services.GetReportReceiverHttpUrl()
+		logger.L().Debug("setting eventReceiverRestURL", helpers.String("url", eventReceiverRestURL))
+	}
 
 	cfg, err := config.LoadConfig("/etc/config")
```
dwertent, line 57: This is confusing, the name of the file is the same name as the directory: `config.GetServiceURLs("/etc/config/services.json")`

main.go
```diff
@@ -43,13 +43,16 @@ func main() {
...
+	var eventReceiverRestURL string
+	if components.Components.ServiceDiscovery.Enabled {
+		services, err := config.GetServiceURLs("/etc/config/services.json")
+		if err != nil {
+			logger.L().Ctx(ctx).Fatal("failed discovering urls", helpers.Error(err))
+		}
 
-	eventReceiverRestURL := services.GetReportReceiverHttpUrl()
-	logger.L().Debug("setting eventReceiverRestURL", helpers.String("url", eventReceiverRestURL))
+		eventReceiverRestURL := services.GetReportReceiverHttpUrl()
```
amirmalka, line 53: you are shadowing the outer `eventReceiverRestURL` var

mainhandler/kubescapehandler.go
```diff
@@ -232,9 +232,11 @@ func (actionHandler *ActionHandler) kubescapeScan(ctx context.Context) error {
 		scanID:   response.ID,
 	}
 
-	nextHandledTime := time.Duration(WaitTimeForKubescapeScanResponse * time.Second)
-	commandResponseData := createNewCommandResponseData(KubescapeResponse, HandleKubescapeResponse, data, &nextHandledTime)
-	insertNewCommandResponseData(actionHandler.commandResponseChannel, commandResponseData)
+	if actionHandler.eventReceiverRestURL != "" {
```
dwertent, line 235: How about we have a function like: `IsServiceDiscoveryEnabled()` instead of  having this (`actionHandler.eventReceiverRestURL != ""`) statement all over the code

dwertent (changes requested):

README.md
matthyx: `serviceDiscovery` isn't part of `ClusterConfig` so I cannot put it inside, however I can create the file `services.json`

main.go
matthyx: f*** thanks for the eagle eye... I wonder of this worked!

main.go
matthyx: but this is how viper works... you don't specify the filename, but only the dir

mainhandler/kubescapehandler.go
matthyx: I'd rather have the mock reporter, because it's not about having service discovery

README.md
matthyx: I think I should add all 4 files...

mainhandler/kubescapehandler.go
matthyx: I did it with a private field

dwertent (approved):

PR #169, opened 2023-09-04
bump backend module

matthyx:


matthyx:
I will include this directly into https://github.com/kubescape/operator/pull/168

PR #174, opened 2023-09-27
feat(scan): deduplicate events sent to be processed

vladklokun:
## PR Type:
Enhancement

___
## PR Description:
This PR introduces a feature to deduplicate events sent for processing to handlers for Continuous Scanning. It adds a cooldown queue that drops duplicate events during a specified interval. This interval can be configured using the 'eventDeduplicationInterval' parameter in the configuration file. The default interval is set to 2 minutes. The PR also includes relevant tests and updates to the configuration files.

Please keep in mind, to help with review this PR includes only interactive Workload Scan resources cleanup, but not in the background cleanup. Although background cleanup comes at a later time, this PR is sufficient to have a working Operator that support the Continuous Scanning capability.

___
## PR Main Files Walkthrough:

`config/config.go`: Added a new configuration parameter 'EventDeduplicationInterval' to set the interval for event deduplication. The default value is set to 2 minutes.
`config/config_test.go`: Updated the test cases to include the new 'EventDeduplicationInterval' configuration parameter.
`continuousscanning/cooldownqueue.go`: Introduced a new file that implements the cooldown queue for event deduplication. The queue forwards the event to its output channel and starts a cooldown for this event. If the same event is put into the queue during the cooldown, it is silently dropped.
`continuousscanning/service.go`: Replaced the existing eventQueue with the new cooldownQueue. Updated the 'listen' and 'work' methods to use the new queue.
`continuousscanning/service_test.go`: Updated the test cases to use the new cooldownQueue. Also, updated the 'makePod' function to include a unique identifier (UID) for each pod.
`main.go`: Updated the 'SetupContinuousScanning' method call to include the queue size and event cooldown interval parameters.
`mainhandler/handlerequests.go`: Updated the 'SetupContinuousScanning' method to accept queue size and event cooldown interval parameters. These parameters are used when creating the new ContinuousScanningService.
`go.mod`: Added a new dependency 'github.com/hashicorp/golang-lru/v2' which provides an implementation of a Least Recently Used (LRU) cache. This is used in the cooldown queue implementation.
`go.sum`: Updated the checksums for the new and updated dependencies.
`continuousscanning/gvks.py`: Introduced a new Python script to list unique GroupVersionKinds (GVKs) used in Kubernetes resources. This is not directly related to the main feature of this PR but seems to be a utility for the developers.
</details>

___
## Overview
Added a cooldown queue that deduplicates events sent for processing to handlers for Continuous Scanning

**Please open the PR against the `dev` branch (Unless the PR contains only documentation changes)**


continuousscanning/gvks.py
```diff
@@ -0,0 +1,32 @@
+#! /usr/bin/env python3
```
dwertent, line 1: What is this file?

continuousscanning/cooldownqueue.go
```diff
@@ -0,0 +1,81 @@
...
+
+}
+
+func makeEventKey(e watch.Event) string {
+	object, ok := e.Object.(*unstructured.Unstructured)
+	if !ok {
+		return ""
+	}
+
+	eventKey := string(e.Type) + "-" + string(object.GetUID())
```
dwertent, line 56: I think we can remove the type
```suggestion
	eventKey := string(object.GetUID())
```

continuousscanning/cooldownqueue.go
```diff
@@ -0,0 +1,81 @@
...
+	eventKey := makeEventKey(e)
+	logger.L().Debug("Adding event to queue", helpers.String("eventKey", eventKey))
+
+	_, exists := q.seenEvents.Get(eventKey)
+	if exists {
+		logger.L().Debug("key exists, dropping event", helpers.Interface("eventKey", eventKey))
+		return
+	}
+
+	go func() {
```
dwertent, line 71: Why are we starting a new go routine every time? In my experience, it can use a large amount of memory.

main.go
```diff
@@ -123,7 +123,7 @@ func main() {
 
 	if operatorConfig.ContinuousScanEnabled() {
 		go func(mh *mainhandler.MainHandler) {
-			err := mh.SetupContinuousScanning(ctx)
+			err := mh.SetupContinuousScanning(ctx, 512, cfg.EventDeduplicationInterval)
```
dwertent, line 126:
```suggestion
			err := mh.SetupContinuousScanning(ctx, mh.DefaultQueueSize, cfg.EventDeduplicationInterval)
```

mainhandler/handlerequests.go
```diff
@@ -80,12 +81,27 @@ func NewMainHandler(config config.IConfig, k8sAPI *k8sinterface.KubernetesApi) *
 		sendReport:             config.EventReceiverURL() != "",
 	}
 	pool, _ := ants.NewPoolWithFunc(config.ConcurrencyWorkers(), func(i interface{}) {
+		logger.L().Debug("ants pool received job", helpers.Interface("job", i))
```
dwertent, line 84: Are the logs in this function leffovers from debugging?

mainhandler/handlerequests.go
```diff
@@ -115,8 +136,16 @@ func (mainHandler *MainHandler) SetupContinuousScanning(ctx context.Context) err
 	}
 
 	fetcher := cs.NewFileFetcher(rulesReader)
+	if fetcher == nil {
+		panic("fetcher is nil")
```
dwertent, line 140: We should return the error and handle it in the calling function.

utils/reportsender.go
```diff
@@ -0,0 +1,62 @@
+package utils
```
dwertent, line 1: I think you can remove this file

continuousscanning/handlers.go
```diff
@@ -0,0 +1,210 @@
...
+	objRaw, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	return &unstructured.Unstructured{Object: objRaw}, nil
+}
+
+func triggerScan(ctx context.Context, wp *ants.PoolWithFunc, clusterConfig config.IConfig, command *armoapi.Command) error {
+	go utils.AddCommandToChannel(ctx, clusterConfig, command, wp)
```
dwertent, line 80: Do we need a new go routine here?

dwertent (changes requested):
Please see my comments.

Also, it seems like we will trigger scanning/delete when a pod is created/deleted. We should not trigger in such cases, only if the pod/replicaset/job does not have a parent as [implemented in this function](https://github.com/kubescape/k8s-interface/blob/main/k8sinterface/k8sdynamic.go#L219) (you might need to implement something similar for unstructured.Unstructured.

continuousscanning/cooldownqueue.go
vladklokun: No, we cannot.

If we create a pod and then delete it, the cooldown queue will discard the DELETED event. This will prevent the handler that removes CRDs from triggering a Kubescape deleted-resource scan.

continuousscanning/gvks.py
vladklokun: It generates a list of GVKs that we should monitor. I debated whether to include it in the Continuous Scanning tree or not.

continuousscanning/cooldownqueue.go
vladklokun: Because writes to channels can block. In order to not block and stop everything from proceeding, we run the write in a goroutine so writes get some breathing room.

vladklokun: Spawning a lot of goroutines can use a large amount of memory when goroutines are leaking. If they aren’t, there is no problem. And if there are, we just need to make sure they stop leaking by using proper context handlilng.

In other cases, it should not negatively affect the memory consumptions. Go programs are okay with running multiple thousands of goroutines. There’s no reason not to use them.

mainhandler/handlerequests.go
vladklokun: No, I consciously added these debug logs in cases where we would need to debug them ever again. They are debug-level, so they will not show up during normal operation.

mainhandler/handlerequests.go
vladklokun: Now _that_ is a leftover from debugging. Will fix.

utils/reportsender.go
vladklokun: Good catch! That was left over from the merge.

continuousscanning/handlers.go
vladklokun: Here we can skip it.

mainhandler/handlerequests.go
vladklokun: Fixed.

utils/reportsender.go
vladklokun: Fixed.

main.go
vladklokun: Fixed.

continuousscanning/gvks.py
dwertent: You could have it in this tree, but maybe have it nested under a "test" directory or something so it will be clear why it is here.

continuousscanning/cooldownqueue.go
dwertent: Ok for now, we should monitor this carefully

dwertent (dismissed):
Well done :)

vladklokun:
I added a fix to skip triggering orphan workloads.

dwertent (approved):

PR #180, opened 2023-10-24
VEX support

slashben:
## PR Type:
Enhancement, Other

___
## PR Description:
This PR introduces the following changes:
- Adds support for VEX (Vulnerability Exchange) by implementing cleanup of VEX documents in storage.
- Updates the 'go-cmp' dependency from v0.5.9 to v0.6.0.
- Replaces the 'kubescape/storage' dependency with a specific version.

___
## PR Main Files Walkthrough:

`watcher/watcher.go`: Added a check to delete OpenVulnerabilityExchangeContainers if the instance ID is not recognized. This is part of the VEX support implementation.
`go.mod`: Updated the 'go-cmp' dependency version and replaced the 'kubescape/storage' dependency with a specific version.
`go.sum`: Updated the checksums for the updated and replaced dependencies.
</details>

___
## Overview

Adding support for VEX: cleanup of VEX document in storage. Part of this feature https://github.com/kubescape/kubevuln/issues/155

- [X] Yes, I signed my commits.


slashben:
> @slashben is this ready?

Still waiting for @vladklokun to review

watcher/watcher.go
```diff
@@ -282,6 +282,18 @@ func (wh *WatchHandler) HandleSBOMFilteredEvents(sfEvents <-chan watch.Event, pr
 			continue
 		}
 
+		if !slices.Contains(wh.managedInstanceIDSlugs, hashedInstanceID) {
+			wh.storageClient.SpdxV1beta1().OpenVulnerabilityExchangeContainers(obj.ObjectMeta.Namespace).Delete(context.TODO(), obj.ObjectMeta.Name, v1.DeleteOptions{})
```
dwertent, line 286: Run this flow only when the VEX capability is enabled

dwertent (changes requested):

watcher/watcher.go
slashben: If VEX capability is not enabled then there are no VEX objects to delete. I don't want to pollute the code by adding `if`-s all around these components. The nice thing that the whole system is object driven, if the object not there nothing happens

dwertent:
We manage the delete in the storage

PR #181, opened 2023-10-24
fix: use a dummy report sender with no reporter URL

vladklokun:
## PR Type:
Bug fix

___
## PR Description:
This PR introduces a dummy report sender to prevent panic when trying to send events with no server URL provided. The dummy report sender stands in for a proper report sender but does not report anywhere. It is used when the EventReceiverURL in the configuration is empty. The PR also includes an update to the 'kubescape/backend' dependency.

___
## PR Main Files Walkthrough:

`utils/reportsender.go`: Introduced a new struct 'dummyReportSender' with methods that mimic a real report sender but do not perform any actions. This is used as a placeholder when no server URL is provided.
`utils/typesutils.go`: Modified the 'NewSessionObj' function to use the dummy report sender when the EventReceiverURL in the configuration is empty.
`go.mod`: Updated the 'kubescape/backend' dependency from v0.0.10 to v0.0.11.
`go.sum`: Updated the checksums related to the 'kubescape/backend' dependency.
</details>

___
## Overview
This PR fixes a panic when trying to send events with no server URL provided.

- [ ] If it is a core feature, I have added thorough tests.


utils/reportsender.go
```diff
@@ -0,0 +1,62 @@
+package utils
+
+import (
+	"time"
+)
+
+type dummyReportSender struct{}
```
dwertent, line 7: This code should be in the `backend` package

dwertent (changes requested):

utils/reportsender.go
vladklokun: https://github.com/kubescape/backend/pull/14

vladklokun: We merged the PR.

vladklokun:
@dwertent 
We should also create a release for this fix, right? I added the label.

dwertent (approved):

PR #212, opened 2024-03-03
send skipTLSVerify and Insecure in image scanning command

amirmalka:
## Overview
This PR fixes https://github.com/kubescape/kubevuln/issues/209

When scanning individual images the operator did not send the `skipTLSVerify` and `http` attributes (for insecure registries). These attributes are supported in the registry scan secret 
(see  https://kubescape.io/docs/operator/vulnerabilities/#granting-credentials-directly )

As part of the fix, when parsing the secret, we match the registry name of the scanned image with the registry name in the secret. If such match found, we pass the values to the scan command args.

When scanning image registry, the scan command did include these fields so no change is needed.

### DIFF
--- a/mainhandler/imageregistryhandler.go
+++ b/mainhandler/imageregistryhandler.go
@@ -800,7 +800,13 @@ func (registryScan *registryScan) setRegistryInfoFromConfigMap(registryInfo *arm
 	registryInfo.Exclude = registryConfig.Exclude
 }
 
-func getRegistryScanSecrets(k8sAPI *k8sinterface.KubernetesApi, namespace, secretName string) ([]k8sinterface.IWorkload, error) {
+// KubernetesApiSecrets is an interface for getting workloads from k8s api
+type IWorkloadsGetter interface {
+	GetWorkload(namespace, kind, name string) (k8sinterface.IWorkload, error)
+	ListWorkloads2(namespace, kind string) ([]k8sinterface.IWorkload, error)
+}
+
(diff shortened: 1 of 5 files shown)


mainhandler/vulnscan.go
```diff
@@ -399,17 +427,19 @@ func (actionHandler *ActionHandler) getAuthConfig(pod *corev1.Pod, imageTag stri
...
 
-	// TODO: this should not happen every scan
-	// build a list of secrets from the the registry secrets
-	secrets, err := cloudsupport.GetImageRegistryCredentials(imageTag, pod)
-	if err != nil {
-		return nil, err
-	}
-	for i := range secrets {
-		registryAuth = append(registryAuth, secrets[i])
+	if pod != nil {
```
amirmalka, line 430: this check is needed go make the code testable (since `GetImageRegistryCredentials` calls `NewKubernetesApi`

mainhandler/imageregistryhandler.go
```diff
@@ -800,7 +800,13 @@ func (registryScan *registryScan) setRegistryInfoFromConfigMap(registryInfo *arm
 	registryInfo.Exclude = registryConfig.Exclude
 }
 
-func getRegistryScanSecrets(k8sAPI *k8sinterface.KubernetesApi, namespace, secretName string) ([]k8sinterface.IWorkload, error) {
+// KubernetesApiSecrets is an interface for getting workloads from k8s api
+type IWorkloadsGetter interface {
```
dwertent, line 804: Is this in use?

amirmalka: Yes, to make the code testable. see `vulnscan_test.go`

dwertent (approved):

PR #230, opened 2024-07-09
add ns include/exclude config for scans

matthyx:


continuousscanning/watchbuilder.go
```diff
@@ -17,7 +17,7 @@ func NewDynamicWatch(ctx context.Context, client dynamic.Interface, gvr schema.G
 	var w watch.Interface
 	var err error
 	if k8sinterface.IsNamespaceScope(&gvr) {
-		w, err = client.Resource(gvr).Namespace("").Watch(ctx, opts)
+		w, err = client.Resource(gvr).Namespace("").Watch(ctx, opts) // TODO support ExcludeNamespaces and IncludeNamespaces
```
matthyx, line 20: this will be part of finishing continuous scanning

Bezbran (approved):

PR #239, opened 2024-08-21
update to go 1.23 and base image to debian12

matthyx:


go.mod
idohuber, line 1: its part of the release or part of local testing?

idohuber (approved):

PR #278, opened 2024-12-25
fix: scan complete application profiles

amirmalka:
## Overview

- Change instance ID field in image scan command to use slug (reverts changes in bf8a747)
- Scan application profiles only for complete ones (not partial)

### DIFF
--- a/mainhandler/vulnscan.go
+++ b/mainhandler/vulnscan.go
@@ -498,6 +498,10 @@ func (actionHandler *ActionHandler) scanApplicationProfile(ctx context.Context, 
 		},
 	}
 
+	if actionHandler.reporter != nil {
+		prepareSessionChain(sessionObj, cmd, actionHandler)
+	}
+
 	if err := sendCommandToScanner(ctx, actionHandler.config, cmd, apis.TypeScanApplicationProfile); err != nil {
 		return fmt.Errorf("failed to send command to scanner with err %v", err)
 	}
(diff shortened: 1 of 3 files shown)


mainhandler/vulnscan.go
```diff
@@ -498,6 +498,10 @@ func (actionHandler *ActionHandler) scanApplicationProfile(ctx context.Context,
 		},
 	}
 
+	if actionHandler.reporter != nil {
+		prepareSessionChain(sessionObj, cmd, actionHandler)
```
amirmalka, line 502: restores missing job IDs

mainhandler/vulnscan.go
```diff
@@ -532,9 +536,9 @@ func (actionHandler *ActionHandler) getImageScanCommand(containerData *utils.Con
 		cmd.Args[identifiers.AttributeUseHTTP] = true
 	}
 
-	// Add instanceID only if not empty
-	if containerData.InstanceID != "" {
-		cmd.InstanceID = &containerData.InstanceID
+	// Add instanceID only if container is not empty
+	if containerData.Slug != "" {
```
amirmalka, line 540: Restores the behaviour that was changed in https://github.com/kubescape/operator/commit/bf8a74795b7e163eaf76c655a86f91d40ce62bc4

amirmalka: It used to be like this: https://github.com/kubescape/operator/blob/v0.2.49/mainhandler/vulnscan.go#L517

matthyx (approved):

PR #279, opened 2025-01-06
use a cooldown queue for AP watcher

matthyx:


watcher/applicationprofilewatcher.go
```diff
@@ -19,7 +19,7 @@ import (
 
 // ApplicationProfileWatch watches and processes changes on ApplicationProfile resources
 func (wh *WatchHandler) ApplicationProfileWatch(ctx context.Context, workerPool *ants.PoolWithFunc) {
-	inputEvents := make(chan watch.Event)
+	eventQueue := NewCooldownQueue()
```
matthyx, line 22: question: is the default 5s enough for AP?

amirmalka: I would increase it to 10s-15s to be on the safe side

matthyx:
@amirmalka I've increased the delay

amirmalka (approved):

amirmalka:
@matthyx I've added the release label in case we want to have a new helm chart

PR #285, opened 2025-01-21
DO NOT MERGE add registry logs

refaelm92:
Signed-off-by: refaelm <refaelm@armosec.io>

## Overview
This PR fixes #

- [ ] Yes, I signed my commits.


watcher/registryhandler.go
```diff
@@ -114,19 +117,22 @@ func (ch *RegistryCommandsHandler) Start() {
 			} else if len(payload) > 0 {
 				status.Payload = payload
 			}
-			ch.patchCommandStatus(&cmd, status)
 
+			logger.L().Ctx(ctx).Info(fmt.Sprintf("finished handling %s command ", cmd.Spec.CommandType))
```
matthyx, line 121: actually we only use .Ctx(ctx) when we want to enrich the open telemetry span (and the ingestion is broken in our backend for the moment) can you remove them unless you log in Warning, Error or Fatal ?

PR #298, opened 2025-03-11
add pullsecrets from pod for scanAP

matthyx:


mainhandler/vulnscan.go
```diff
@@ -253,6 +248,21 @@ func (actionHandler *ActionHandler) scanApplicationProfile(ctx context.Context)
 		return errors.New("kubevuln is not enabled")
 	}
 
+	// get the pod from the session object
```
matthyx, line 251: relevant code here

PR #313, opened 2025-05-16
fix: private container registry credentials are not used when scanning an application profile

janeisklar:
## Overview

When the operator processes an `ApplicationProfile` event and sends a `scanApplicationProfile` command the following arguments are supplied:
[arguments](https://github.com/kubescape/operator/blob/d69efef8bd4a384dfafadf82ee21ca1d0a72e1f6/watcher/applicationprofilewatcher.go#L110-L113)

When that command is then executed, credentials are not passed on because the `pod` argument is missing:
[obtaining credentials](https://github.com/kubescape/operator/blob/d69efef8bd4a384dfafadf82ee21ca1d0a72e1f6/mainhandler/vulnscan.go#L257-L262)

## Solution

The pod is loaded and passed so that the credentials are loaded.

## Additional Information

We observed this issue with pods that are started by Argo Workflows. They only run for a very limited time and thus were not scanned by other means.

I also considered simply passing the imageID as a new argument (because then we wouldn't need to load the whole Pod which may not even exist anymore). This would have been sufficient for my use-case (Azure CR) but then I saw that this wouldn't work for other credential extraction methods like imagePullSecrets - so it's probably not worth it.

put an [x] in the box to get it checked 

### DIFF
--- a/watcher/applicationprofilewatcher.go
+++ b/watcher/applicationprofilewatcher.go
@@ -2,6 +2,7 @@ package watcher
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/armosec/armoapi-go/apis"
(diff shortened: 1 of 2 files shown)


matthyx:
@janeisklar so this is only valid for single Pods created by Argo CD or similar, maybe we should consider adding other types as well and inspect their templates to find the projected Pod spec?

However, I don't understand why the SBOM isn't already created by the node-agents... can you check if you have nodeSbomGeneration enabled?

matthyx (approved):

matthyx:
thanks @janeisklar I think that's the best we can do here, regardless of the SBOM being generated by node-agent or not

PR #374, opened 2026-05-20
CRD-driven CEL admission rules with async validator

slashben:
Implements [designs-and-proposals#3 (CRD-driven CEL admission rules)](https://github.com/kubescape/designs-and-proposals/pull/3) (merged).

## Summary

Replaces the operator's hardcoded admission rules (R2000 Exec, R2001 Port-Forward) with CRD-defined CEL rules. Same `Rules` CRD schema used by the node-agent for eBPF events, extended with a new `k8s-admission` event type.

### Core pipeline
- **`admission/cel/`** — `AdmissionCEL` engine + `AdmissionCelEvent` adapter (exposes `event.Kind`, `event.UserInfo`, etc. to CEL; `object`/`oldObject`/`options` as top-level variables due to `ext.NativeTypes()` map limitation).
- **`admission/rules/cel/`** — `CelRuleCreator` and `CelRuleEvaluator` replace the legacy Go rule descriptors and factory.
- **`admission/ruleswatcher/`** — Dynamic informer over `Rules` CRDs; filters for enabled rules with at least one `k8s-admission` expression, calls `CelRuleCreator.SyncRules`.
- **`admission/rulebinding/cache/`** — Now takes the `RuleCreator` as a constructor argument and exposes `RefreshRules()` for the watcher to invoke.

### Performance and safety (per the design doc)
- **Self-pod short-circuit** — Validator parses the unverified `sub` claim from the projected ServiceAccount token at startup and drops any admission request whose `UserInfo.Name` matches. Hard guarantee against feedback loops from the operator's own writes. Falls back gracefully if the token isn't readable.
- **Kind-of-interest pre-filter** — `CelRuleCreator` extracts `event.Kind == "X"` patterns from loaded expressions at `SyncRules` time. Validator skips events for Kinds no rule could match. Conservative — falls back to wildcard for expressions with `||` or no Kind pin.
- **Async validator with bounded worker pool** — `Validate()` enqueues onto a 1000-slot buffered channel and returns immediately; 10 workers drain it. The API server never waits on CEL. Queue-full → drop with throttled warning log + atomic counter (exposed via `DropCount()` for Prometheus).

### Bug fix
- **`admission/rulebinding/cache/`** — Both the RBCache and RulesWatcher adaptors share the dynamic watcher event stream. The RBCache now filters non-`RuntimeRuleAlertBinding` events before parsing, fixing the `"cannot convert int64 to string"` spam previously seen after adding the RulesWatcher.

### Deleted
- `admission/rules/v1/r2000_exec_to_pod.go` (+ test)
- `admission/rules/v1/r2001_portforward.go` (+ test)
- `admission/rules/v1/factory.go`
- `admission/rules/v1/rule.go`

Kept: `helpers.go` (K8s enrichment functions), `failureobject.go` (`GenericRuleFailure`), `rule_interface.go`.

## Validation

End-to-end on a kind cluster with the locally built image:
- `kubectl exec` → R2000 fires, alert in operator logs with K8s enrichment populated
- `kubectl port-forward` → R2001 fires
- `kubectl apply` NetworkPolicy → R2002 fires (additional rule from the kind test), `event.UserInfo.Username` resolves correctly in the message expression
- Pod CREATE with no matching rule → no alert (negative test)
- Operator self-generated request → short-circuited
- 309 unit tests pass

## Dependencies

| Dep | Status |
|-----|--------|
| `armoapi-go` `EventTypeK8sAdmission` constant | ✅ merged, v0.0.714 |
| `rulelibrary` R2000/R2001 as CEL YAML | TODO — separate PR |
| `helm-charts` updates (RBAC, webhook config, image bump) | TODO — separate PR, marked DO NOT MERGE |

## Test plan

- [x] Unit tests: 309 pass on `go test -race ./...`
- [x] Kind-cluster e2e: exec, port-forward, NetworkPolicy creates, negative pod-create, self-request short-circuit
- [ ] Reviewer feedback on the design doc and architecture

## Release Notes

* **New Features**
  * Introduced a CEL-based admission rules engine driven by Kubernetes Custom Resources, including rule pre-filtering by resource kind and per-rule evaluation caching.
  * Added an asynchronous admission evaluation pipeline (bounded worker pool) with queue-drop tracking.
  * Added support for dynamic rule syncing via a CRD watcher and a self-request short-circuit to reduce redundant evaluations.
* **Refactor / Change**
  * Migrated pod exec and port-forward admission logic to CEL-based rules; removed the legacy built-in implementations.
* **Documentation**
  * Added a feature guide for CEL admission rules.


admission/webhook/validator.go
```diff
@@ -17,54 +19,245 @@ import (
...
+	if ui == nil {
+		return false
+	}
+	return ui.GetName() == av.selfSubject
+}
+
+// Validate implements admission.ValidationInterface. The API server is
+// synchronous from the framework's perspective, but our rule evaluation is
+// not: matching requests are enqueued onto a bounded worker pool and the
+// validator returns nil immediately. The API server never waits on CEL.
```
matthyx, line 181: sorry I don't understand how that works... does it mean we are never denying requests (since we always return nil)?

admission/rules/cel/creator.go
```diff
@@ -0,0 +1,136 @@
...
+	celEngine  *admissioncel.AdmissionCEL
+}
+
+var _ rules.RuleCreator = (*CelRuleCreator)(nil)
+
+// NewCelRuleCreator returns a new CelRuleCreator with no rules loaded yet.
+func NewCelRuleCreator(celEngine *admissioncel.AdmissionCEL) *CelRuleCreator {
+	return &CelRuleCreator{
+		celEngine: celEngine,
+		// Empty creator accepts nothing — the kind filter starts non-wildcard
```
matthyx, line 26: is it fine to drop all admissions until we have rules? maybe we should only accept HTTP connections after we have rules?

admission/rules/cel/evaluator.go
```diff
@@ -0,0 +1,212 @@
...
+		},
+		RuleAlert: apitypes.RuleAlert{
+			RuleDescription: e.rule.Description,
+		},
+		AdmissionAlert: buildAdmissionAlert(attrs),
+		RuleID:         e.rule.ID,
+	}
+
+	// Enrich with K8s details when a cache is available.
+	if access != nil {
```
matthyx: enrichK8sDetails calls rulesv1.GetControllerDetails(attrs, clientset) which does a live clientset.CoreV1().Pods(namespace).Get(...), it only makes sense for a Pod or a Pod subresource.

```suggestion
	if access != nil && (attrs.GetResource().Resource == "pods" || attrs.GetKind().Kind == "PodExecOptions") {
```

admission/cel/cel.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/google/cel-go/ext"
+)
+
+// AdmissionCEL owns a CEL environment configured for evaluating expressions
+// against AdmissionCelEvent values. Compiled programs are cached so repeated
+// evaluation of the same expression avoids re-compilation.
+type AdmissionCEL struct {
+	env          *cel.Env
+	programCache map[string]cel.Program
```
matthyx: cache is never cleared if some rules are dropped

admission/webhook/validator.go
```diff
@@ -17,54 +19,245 @@ import (
...
 	return nil
 }
 
+// runWorker drains the evaluation queue until ctx is canceled.
+func (av *AdmissionValidator) runWorker(ctx context.Context) {
+	defer av.wg.Done()
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
matthyx, line 226: you should probably drain the jobs channel before exit, other jobs enqueued just before context cancellation will never get processed

matthyx (changes requested):
_you have to explain to me how the async pipeline blocks admission, because now I understand that we never block any admission_
ok by design we never block anything, this is only for alerting

admission/webhook/validator.go
matthyx: ok it's alert only

Bezbran (commented):
I didn't see anything hurting my eyes.
At the beginning I wondered why we are not "blocking" but @matthyx asked and answered.

admission/rules/cel/creator.go
slashben: IMHO - yes, I don't want to hold any line up. What if the user doesn't install rules? The whole design idea is that we treat admission webhooks as best-effort event sources

admission/rules/cel/evaluator.go
slashben: You are right!

admission/webhook/validator.go
slashben: I think you're right

slashben:
Pushed three fixes addressing @matthyx's review:

| # | Issue | Commit |
|---|-------|--------|
| 3 | `enrichK8sDetails` only valid for Pod / pod subresources | a2db526 — added `enrichmentApplicable` gate (Pod, PodExecOptions, PodPortForwardOptions, PodAttachOptions). Other kinds skip enrichment instead of fetching an unrelated Pod by name. |
| 4 | CEL program cache leaks for removed rules | d3af050 — added `AdmissionCEL.RetainOnly(activeExpressions)`. `CelRuleCreator.SyncRules` now collects every expression string (RuleExpression + Message + UniqueID) and prunes the cache after each swap. |
| 5 | Drain queued jobs on shutdown | c2920a8 — on `ctx.Done()` the worker switches to a bounded (10s) non-blocking drain pass before exiting. If the drain deadline trips, remaining jobs are logged with a count and abandoned. |

On the "block admissions until rules are loaded" question (creator.go:26) — I'd like to keep the current behavior. The operator is an alert-only audit path, the events are best-effort information, and gating admission on rule readiness would couple API server availability to operator startup. Cheaper to drop the cold-start events than to add a sync+blocking dance.

Tests for all three fixes are in this push. `go test -race ./...` is green (322 tests across 21 packages).

admission/rules/cel/evaluator.go
```diff
@@ -0,0 +1,232 @@
...
+
+// SetParameters stores per-binding parameter overrides.
+func (e *CelRuleEvaluator) SetParameters(parameters map[string]interface{}) {
+	e.parameters = parameters
+}
+
+// GetParameters returns the per-binding parameter overrides.
+func (e *CelRuleEvaluator) GetParameters() map[string]interface{} {
+	return e.parameters
+}
```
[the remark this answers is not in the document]
matthyx: yes I agree with this one

admission/cel/cel.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		matched = true
+
+		out, err := c.evaluateProgram(expr.Expression, evalContext)
+		if err != nil {
+			return false, err
+		}
+		// nil means the program was previously cached as a compile failure.
+		if out == nil {
+			return false, nil
```
matthyx: we have to be careful to catch the compilation failure, otherwise we will think this rule never evaluates to false while it's just not compiling

slashben:
Pushed the next round of fixes (branch force-pushed after rebase onto current `main`):

| Issue | Commit | Notes |
|-------|--------|-------|
| Cache compile failures as nil silently swallows broken rules (matthyx cel.go:100) | 4d2cce5 | Cache value is now `cachedProgram{prog, err}`; every lookup re-returns the cached error. The evaluator already logs and returns nil-match on errors, so a broken expression now produces a recurring loud warning instead of a silent no-op. New test verifies the error is surfaced on every call. |
| Per-binding parameters never reach CEL evaluation (matthyx + coderabbit evaluator.go:55) | 98b0398 | Declared `params` as a top-level CEL variable; `ProcessEvent` injects `e.parameters` under that key. CEL expressions can now reference `params["allowedNamespaces"]`, `params["threshold"]`, etc. New parameterized rule test confirms the override path. |
| CodeQL overflow alert on `len(rs)*3` (creator.go:57) | 4fd9434 | Dropped the capacity hint; Go's slice growth handles the small per-rule expression count. |
| Branch was conflicting with `main` | (rebase) | Rebased onto current `main`; conflict was on the legacy R2000/R2001 v1 files that we delete intentionally. |

`go test -race ./...` is green: 335 tests, 21 packages.

Not addressed in this push:
- CodeRabbit's "give cancellation priority before consuming more jobs" — the `ctx` we'd pass to `evaluate` is the worker context, not the drain context, so a job picked just after cancellation evaluates against an already-canceled context. The drain path uses a fresh bounded context. I think the current behavior is correct, but happy to add the explicit `ctx.Err()` check if you'd like.
- CodeRabbit's nil-`ruleSyncer` guard in `NewRulesWatcher` — main.go always constructs with a non-nil syncer; adding a panic guard would mask a wiring bug rather than catch it.
- Markdown lint on the architecture fence in `docs/features/cel-admission-rules.md` — happy to fix in a followup if it blocks merge.

matthyx (approved):
looks good now

PR #379, opened 2026-05-27
Fix empty-Wlid scan dispatch in operator SBOM watcher

### LINKED ISSUE #378: Operator SBOM watcher dispatches `TypeScanImages` commands with empty `Wlid` when SBOM event beats Pod event, vuln results lose workload attribution
# Description

The operator's `SBOMWatcher` correlates incoming `SBOMSyft` events with their owning workload by looking up the image hash in an in-memory map (`ImageToContainerData`) that is populated by Pod events from a separate informer channel. The two channels are merged in a `select`:

```go
// operator/watcher/sbomwatcher.go:81-104 (excerpt)
for {
    select {
    // FIXME select processes the events randomly, so we might see the SBOM event before the pod event
    case event := <-wh.eventQueue.ResultChan: // pod events fill ImageToContainerData
        ...
        for _, containerStatus := range containerStatuses {
            hash := hashFromImageID(containerStatus.ImageID)
            wh.ImageToContainerData.Set(hash, utils.ContainerData{
                ContainerName: containerStatus.Name,
                Wlid:          wlid,
            })
        }
    case sbomEvent, ok := <-sbomEvents:
        if ok { eventQueue.Enqueue(sbomEvent) } else { ... }
    ...
}
```

Two problems compose:

1. **Initial list races with pod cache fill.** Before the `select` loop starts, the watcher pages the *full existing* SBOM list and enqueues each as `watch.Added` (`sbomwatcher.go:65-77`). The race is sharper than the FIXME at line 82 suggests, because `HandleSBOMEvents` runs as its own goroutine (started at `sbomwatcher.go:54`, *before* the SBOM list call) and consumes from a separate cooldown queue. The `ImageToContainerData` map it reads from is only populated by the main `select` loop draining pod events from `wh.eventQueue` (lines 83-101). So even though `listPods` is invoked synchronously at line 45 when `ServiceDiscovery.Enabled == true`, it only **enqueues** pod events  it does not directly populate the map. The map is filled lazily by the select loop, which has not yet started when the SBOM goroutine begins servicing events.

   Two concrete failure paths:
   - With `ServiceDiscovery.Enabled == true`: race between the SBOM-handling goroutine and the select loop draining queued pod events. Whichever wins decides whether `Wlid` is empty.
   - With `ServiceDiscovery.Enabled == false` (line 43): `listPods` never runs at all, so `ImageToContainerData` is never populated and **every** SBOM dispatch goes out with `Wlid=""`.

2. **`validateContainerData` does not check `Wlid`.** When `HandleSBOMEvents` processes an SBOM whose image hash isn't in `ImageToContainerData` yet, `imageContainerData` is the zero-value struct and `containerData.Wlid == ""`. The validation step only rejects empty `ImageID`/`ImageTag`:

   ```go
   // sbomwatcher.go:221-229
   func validateContainerData(containerData *utils.ContainerData) error {
       if containerData.ImageID == "" { return ErrMissingImageID }
       if containerData.ImageTag == "" { return ErrMissingImageTag }
       return nil
   }
   ```

   So validation passes, and a `TypeScanImages` command is dispatched with `Wlid = ""`:

   ```go
   // sbomwatcher.go:181-191
   cmd := &apis.Command{
       Wlid:        containerData.Wlid,          // = ""
       CommandName: apis.TypeScanImages,
       Args: map[string]interface{}{
           utils.ArgsContainerData: containerData,
       },
   }
   logger.L().Info("scanning SBOM", helpers.String("wlid", cmd.Wlid), ...)
   producedCommands <- cmd
   ```

   The downstream `kubevuln` HTTP scan endpoint (`controllers/http.go:140-175`) accepts an empty `Wlid` and scans the image. `ScanService.ValidateScanCVE` (`kubevuln/core/services/scan.go:857-881`) only checks `ImageHash` and `ImageSlug`, so the empty-`Wlid` command passes validation and the scan executes. Then at `kubevuln/core/services/scan.go:499-507`, the platform CVE submission is gated on `if workload.Wlid != ""` — meaning the manifest is computed but **silently never submitted** to the backend when `Wlid` is empty. End-to-end effect: the operator dispatches a malformed scan, kubevuln burns the CPU/network to scan the image, and the result is silently dropped from the platform submission path. There is no retry path: once the SBOM event has been drained from the channel, it is gone, and no error log above `info` is emitted anywhere along the way.

   The presence of the `if workload.Wlid != ""` guard in kubevuln suggests the empty-`Wlid` case was already a known-bad input on the consumer side; it was patched defensively downstream rather than rejected at the source.

# Environment
- Repo: `kubescape/operator` at `main` HEAD
- Go: 1.25.0 (per `go.mod`)
- Tests: `go test ./...` in `operator/` — passes (the race is not covered by an existing test).

# Steps To Reproduce

This is hard to script in a unit test today because `HandleSBOMEvents` and the populating goroutine share state through a private map. Conceptually:

1. Apply a `SBOMSyft` object to storage with valid `ImageID`/`ImageTag` annotations but **before** the corresponding Pod is observed by the operator (e.g. delete the operator pod, leave SBOMs and pods alive, then restart only the operator, the initial SBOM list runs immediately while the pod informer is still warming up).
2. Observe `kubescape scanning SBOM` log lines with `wlid=""`.
3. Inspect the vuln report, the affected entries have no workload attribution.

A minimal reproducible unit test could be added by exposing a test seam (e.g. injecting the initial list and pod informer as separate channels in a constructor).

# Expected behavior

Either:

- **Wait for the Wlid.** If the image hash is not yet in `ImageToContainerData`, the SBOM event should be parked (re-enqueued with backoff, or buffered until the corresponding pod event arrives) rather than dispatched with an empty Wlid. The map can also be pre-populated by listing pods *before* the SBOM list call.

- **Reject explicitly.** If parking is too complex, `validateContainerData` should treat empty `Wlid` as an error — at least surfacing the problem as `ErrMissingWlid` instead of dispatching a malformed command. This trades a silent attribution loss for an explicit, retryable failure.

In either case the FIXME at line 82 should be resolved, not just annotated.

# Actual Behavior

- On operator startup, the full existing SBOM list is enqueued before the pod cache has filled. Affected entries are dispatched with `Wlid=""`.
- When `ServiceDiscovery.Enabled == false`, `listPods` is never called and **every** SBOM dispatch goes out with `Wlid=""`.
- `validateContainerData` accepts the empty-`Wlid` case as valid.
- A vuln scan command is sent over the worker pool with the malformed identifier.
- `kubevuln`'s `ValidateScanCVE` does not check `Wlid` either, so the scan executes.
- At `kubevuln/core/services/scan.go:500` the platform CVE submission is gated on `if workload.Wlid != ""` and silently no-ops for empty `Wlid` — the scan result is computed but never submitted to the backend.
- No log line above `info` is emitted that would let an operator know correlation was lost or that the submission was skipped.

# Suggested fix

Minimum, low-risk change in `operator/watcher/sbomwatcher.go`:

```go
// sbomwatcher.go:221
func validateContainerData(containerData *utils.ContainerData) error {
    if containerData.ImageID == "" {
        return ErrMissingImageID
    }
    if containerData.ImageTag == "" {
        return ErrMissingImageTag
    }
    if containerData.Wlid == "" {
        return ErrMissingWlid
    }
    return nil
}
```

Combined with re-queueing the SBOM event on `ErrMissingWlid` (with a small TTL/backoff) so a slightly-too-early arrival doesn't drop the scan permanently. Pseudocode in `HandleSBOMEvents`:

```go
if err := validateContainerData(containerData); err != nil {
    if errors.Is(err, ErrMissingWlid) {
        wh.deferredSBOMs.Enqueue(e, time.Now().Add(5*time.Second))
        continue
    }
    errorCh <- err
    continue
}
```

A small worker drains `deferredSBOMs` on a timer and re-enqueues into the main queue, with a max-retries cap.

A better long-term fix: order the startup explicitly. List pods first, populate `ImageToContainerData` synchronously, *then* call the initial SBOM `EachListItem`. The current code does the SBOM list (`sbomwatcher.go:65`) without waiting on any pod-informer sync signal.

Regression test: inject a fake informer pair into `SBOMWatcher` whose pod channel is blocked at start. Push an SBOM event. Assert that either (a) no command is dispatched with `Wlid=""`, or (b) the SBOM is re-enqueued and eventually dispatched with the correct Wlid after the pod event is unblocked.

# Source
- `operator/watcher/sbomwatcher.go:82` — the FIXME marking the race.
- `operator/watcher/sbomwatcher.go:43-51` — `listPods` only runs when `ServiceDiscovery.Enabled`; even then it only enqueues pod events rather than populating the map directly.
- `operator/watcher/sbomwatcher.go:54` — `HandleSBOMEvents` goroutine started before the SBOM list paging at lines 65-77.
- `operator/watcher/sbomwatcher.go:65-77` — initial SBOM list, enqueued before the pod informer warms up.
- `operator/watcher/sbomwatcher.go:137-191` — `HandleSBOMEvents` reads the racing map (line 163) and dispatches the scan command (lines 181-190).
- `operator/watcher/sbomwatcher.go:221-229` — `validateContainerData` (missing the `Wlid` check).
- `kubevuln/controllers/http.go:140-175` — HTTP handler that forwards the command to the scan service.
- `kubevuln/core/services/scan.go:857-881` — `ValidateScanCVE` only validates `ImageHash` and `ImageSlug`, not `Wlid`.
- `kubevuln/core/services/scan.go:499-507` — `if workload.Wlid != ""` guard that silently skips platform CVE submission when `Wlid` is empty.

yugal07:
## Summary

When an `SBOMSyft` event was processed before the owning Pod had populated `ImageToContainerData` — either during the startup race (initial SBOM list paged before the pod informer warmed up) or unconditionally when `ServiceDiscovery.Enabled == false` — the operator dispatched a `TypeScanImages` command with `Wlid=""`. `kubevuln` then ran the scan but its `if workload.Wlid != ""` guard silently dropped the result from the platform CVE submission path. No log above `info` was emitted, so the lost attribution was invisible to operators.

This PR closes the race at the source: empty `Wlid` is now a validation failure, and SBOMs that hit it are parked and retried with bounded exponential backoff rather than dispatched as malformed scans.

## Changes

**`operator/watcher/watchhandler.go`**
- Added `ErrMissingWlid` (alias of existing `ErrMissingWLID`) for the new validation branch.
- Added `sbomRetryAttempts maps.SafeMap[string, int]` to track per-SBOM retry counts.

**`operator/watcher/sbomwatcher.go`**
- `validateContainerData` now returns `ErrMissingWlid` when `Wlid == ""`.
- `HandleSBOMEvents` treats `ErrMissingWlid` as transient: re-enqueues the SBOM after a delay that doubles each attempt (2s → 4s → 8s → … capped at 60s) for up to 5 attempts, then drops with a `warning` log + `ErrMissingWlid` on `errorCh`. Retry bookkeeping is cleared on success and on exhaustion to avoid leaks.
- Backoff parameters live in package-level `var`s (`sbomRetryInitialDelay`, `sbomRetryMaxDelay`, `sbomRetryMaxAttempts`) so tests can shorten them.

**`operator/watcher/sbomwatcher_test.go`**
- Updated `TestHandleSBOMEvents` to seed `ImageToContainerData` so the existing cases now assert the correct `Wlid` is carried on the command (the prior `expectedCommands` literally encoded the bug — empty `Wlid` on dispatched commands).
- New `TestValidateContainerData` — table-driven coverage of all four validation branches, including the `#378` regression case.
- New `TestSBOMRetryBackoff` — verifies the doubling schedule, the `maxDelay` cap, and that absurd attempt counts (e.g. 64, which would overflow the shift) stay capped instead of returning a non-positive duration.
- New `TestHandleSBOMEvents_WlidArrivesLate` — reproduces the race: SBOM arrives first with empty map; pod data lands mid-retry; asserts the eventual scan command carries the correct `Wlid` and that retry bookkeeping is cleared.
- New `TestHandleSBOMEvents_WlidNeverArrives_ExhaustsRetries` — pod data never arrives; asserts retries are bounded, `ErrMissingWlid` is surfaced exactly once, **no** scan command is ever dispatched, and retry bookkeeping is cleared.

## Fixes 
#378 

## Reproduction

Verified end-to-end with a temporary repro test that uses only the pre-fix API:

| Code state                | Repro outcome                                                | Meaning                                                |
| ------------------------- | ------------------------------------------------------------ | ------------------------------------------------------ |
| **Unfixed** (fix stashed) | PASS — asserts dispatched command has `Wlid=""`              | Bug reproduced: malformed scan dispatched              |
| **Fixed** (this PR)       | FAIL — "no scan command and no error within 2s — handler is parking the event" | SBOM is parked for retry instead of dispatched as malformed |

The repro file was deleted after verification; the two new permanent regression tests above cover the same paths.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./watcher/` clean
- [x] `go test ./watcher/ -count=1` — full watcher suite green (~91s)
- [x] Manually reproduced the bug against unfixed code (repro PASS) and confirmed the fix closes it (repro FAIL, with diagnostic "handler is parking the event")

## Notes for reviewers

- The deeper "list pods first, then SBOMs" startup-ordering fix suggested in the issue is **not** included here, it would require restructuring `SBOMWatch` and waiting on a pod-informer sync signal that doesn't exist today. The current fix is sufficient to eliminate the empty-`Wlid` dispatch and is independently valuable even if a later PR also fixes the ordering. I'd dedicate another few days on that long term solution after my exams :smile:  
- `ErrMissingWlid` is intentionally aliased to the existing `ErrMissingWLID` rather than introduced as a separate error, so callers checking either spelling will work.
- The FIXME at `sbomwatcher.go:82` (which the issue called out specifically) is now correct in spirit: the race still exists at the `select`, but it no longer produces malformed scans, the validator + retry loop absorb it.

## Release Notes

* **New Features**
  * SBOM scanning now handles missing workload identification data gracefully, retrying with exponential backoff instead of immediate failure
  * Maximum retry limits prevent infinite retry loops

* **Tests**
  * Substantially improved test coverage for SBOM event processing, validation, and retry behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/operator/pull/379?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/watcher/sbomwatcher.go
+++ b/watcher/sbomwatcher.go
@@ -2,6 +2,7 @@ package watcher
 
 import (
 	"context"
+	"errors"
 	"slices"
 	"strings"
 	"time"
@@ -18,6 +19,14 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/tools/pager"
+)
+
+// Bounded exponential backoff for SBOM events whose owning pod has not been
+// observed yet. Declared as vars (not consts) so tests can shorten them.
+var (
+	sbomRetryInitialDelay = 2 * time.Second
+	sbomRetryMaxDelay     = 60 * time.Second
+	sbomRetryMaxAttempts  = 5
 )
 
 // SBOMWatch watches and processes changes on SBOMs
@@ -169,6 +178,41 @@ func (wh *WatchHandler) HandleSBOMEvents(eventQueue *CooldownQueue, producedComm
 		}
 
 		if err := validateContainerData(containerData); err != nil {
+			if errors.Is(err, ErrMissingWlid) {
+				// The pod that owns this image hasn't been observed yet
+				// (initial-list race, or pod informer not yet warm). Re-enqueue
+				// with bounded exponential backoff rather than dispatch a scan
+				// command with an empty Wlid — kubevuln silently drops those
+				// from the platform submission path.
+				key := obj.ObjectMeta.Namespace + "/" + obj.ObjectMeta.Name
+				attempt := wh.sbomRetryAttempts.Get(key)
+				if attempt >= sbomRetryMaxAttempts {
+					wh.sbomRetryAttempts.Delete(key)
+					logger.L().Warning("dropping SBOM scan after exhausting retries waiting for Wlid",
+						helpers.String("name", obj.ObjectMeta.Name),
+						helpers.String("namespace", obj.ObjectMeta.Namespace),
+						helpers.String("imageID", containerData.ImageID),
+						helpers.Int("attempts", attempt))
+					errorCh <- err
+					continue
+				}
+				wh.sbomRetryAttempts.Set(key, attempt+1)
+				delay := sbomRetryBackoff(attempt)
+				logger.L().Debug("Wlid not yet known for SBOM, re-enqueueing",
+					helpers.String("name", obj.ObjectMeta.Name),
+					helpers.String("namespace", obj.ObjectMeta.Namespace),
+					helpers.String("imageID", containerData.ImageID),
+					helpers.Int("attempt", attempt+1),
+					helpers.String("delay", delay.String()))
+				go func(ev watch.Event, d time.Duration) {
+					time.Sleep(d)
+					if eventQueue.Closed() {
+						return
+					}
+					eventQueue.Enqueue(ev)
+				}(e, delay)
+				continue
+			}
 			logger.L().Error("failed to get container data from SBOM",
 				helpers.String("name", obj.ObjectMeta.Name),
 				helpers.String("namespace", obj.ObjectMeta.Namespace),
@@ -177,6 +221,8 @@ func (wh *WatchHandler) HandleSBOMEvents(eventQueue *CooldownQueue, producedComm
 			errorCh <- err
 			continue
 		}
+		// success: clear any retry bookkeeping for this SBOM
+		wh.sbomRetryAttempts.Delete(obj.ObjectMeta.Namespace + "/" + obj.ObjectMeta.Name)
 
 		cmd := &apis.Command{
 			Wlid:        containerData.Wlid,
@@ -225,5 +271,18 @@ func validateContainerData(containerData *utils.ContainerData) error {
 	if containerData.ImageTag == "" {
 		return ErrMissingImageTag
 	}
+	if containerData.Wlid == "" {
+		return ErrMissingWlid
+	}
 	return nil
+}
+
+// sbomRetryBackoff returns the delay before retry number `attempt` (0-indexed).
+// Doubles each time, capped at sbomRetryMaxDelay.
+func sbomRetryBackoff(attempt int) time.Duration {
+	d := sbomRetryInitialDelay << attempt
+	if d <= 0 || d > sbomRetryMaxDelay {
+		return sbomRetryMaxDelay
+	}
+	return d
 }
--- a/watcher/sbomwatcher_test.go
+++ b/watcher/sbomwatcher_test.go
@@ -3,6 +3,7 @@ package watcher
 import (
 	"context"
 	_ "embed"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


yugal07:
@coderabbitai review

matthyx (approved):
perfect, thanks @yugal07

PR #383, opened 2026-06-11
feat: handle operatorAction commands  (annotate + dry run) 

yugal07:
# operator: handle `operatorAction` commands — Phase 1 remediation (annotate + dry-run)

**Contract:** [armosec/armoapi-go#655 — `operatorAction` command contract](https://github.com/armosec/armoapi-go/pull/655) (merged, shipped in `v0.0.720`)
**Pairs with:** https://github.com/kubescape/helm-charts/pull/855

## What this PR does

This is **Phase 1** of the merged design: it teaches the in-cluster operator to
execute a `TypeOperatorAction` command end to end, with the lowest-blast-radius
action (`annotate`) and **dry-run as the default**. It proves the
CLI → operator → `OperatorCommand` status pipeline with essentially zero risk,
and lays the extensible `Remediator` framework that `quarantine`/`cordon`
plug into in later phases.

No new transport, no new endpoint: remediation is a new *verb* on the existing
`apis.Command` pipeline that already powers `kubescape operator scan`.

- **`go.mod` / `go.sum`** — bump `armosec/armoapi-go` `v0.0.673` → `v0.0.720` to
  pick up the merged `operatorAction` contract (`TypeOperatorAction`,
  `OperatorActionArgs`, `IsDryRun`, `ToArgs`/`OperatorActionArgsFromMap`).
- **`mainhandler/handlerequests.go`** — wire dispatch in two places:
  `case apis.TypeOperatorAction` in `runCommand` (→ `handleOperatorAction`), and
  in the itemize switch so an action is handled as a single request (it carries
  its own target) rather than fanned out per pod.
- **`mainhandler/actionhandler.go`** *(new)* — `handleOperatorAction`: parses the
  typed args, enforces the Phase-1 safety rails, dispatches to the matching
  `Remediator`, writes the result to the `OperatorCommand` status payload, and
  emits a best-effort Kubernetes Event for audit.
- **`mainhandler/remediators/remediator.go`** *(new)* — the `Remediator` interface
  (`Plan`/`Apply`/`Revert`), the `Target`/`Request`/`Plan`/`Result` types, and
  `NewRegistry` (annotate only for now; new actions are added by extending the map,
  not the pipeline).
- **`mainhandler/remediators/annotate.go`** *(new)* — `AnnotateRemediator` over
  Deployment/StatefulSet/DaemonSet/Pod: a JSON merge patch that sets the
  `kubescape.io/remediated`, `…/remediation-reason`, `…/remediation-finding-ref`
  annotations; `Revert` deletes them (null-merge); server-side dry-run support.
- **`mainhandler/remediators/annotate_test.go`**, **`mainhandler/actionhandler_test.go`**
  *(new)* — 16 unit tests.

## Safe-by-default behaviour

- **Dry-run is the default.** `OperatorActionArgs.IsDryRun()` returns `true`
  unless the caller set `dryRun=false` explicitly (the CLI's `--confirm`), so a
  missing flag can never silently perform a real write.
- **Two-layer dry-run, matching the design.** A dry-run still issues the patch
  with server-side `metav1.DryRunAll` so the change is validated against
  admission controllers, never persisted.
- **Namespace safety rail.** Targets in excluded/protected namespaces
  (`config.SkipNamespace`) are rejected before any client call.
- **No default-install power.** The operator gates real writes on RBAC, which the
  default install does not grant (see the paired helm-charts PR). Without it,
  `annotate` returns a clean `Forbidden` that is recorded on the command status —
  nothing is mutated.
- **Explicit scope fences.** `quarantine`/`cordon` return a clear
  "not implemented yet (later phase)" error; selector-driven (findings) targeting
  returns a clear "phase 2" error; unknown actions error explicitly. Phase 1 acts
  only on an explicit `target`.

## Audit trail

Every action records its `Plan`/`Result` (action, target, dryRun, applied) on the
`OperatorCommand` status `payload` (the GitOps- and Headlamp-friendly channel) and
emits a `KubescapeRemediation` Kubernetes Event on the target's namespace.

## Testing

- `go build ./...` and `go vet ./...` — clean.
- New tests pass (`mainhandler`, `mainhandler/remediators`); dry-run assertions use
  a pass-through reactor to verify the `PatchOptions.DryRun` actually sent —
  necessary because the fake clientset's tracker ignores server-side dry-run and
  always persists, so "did not persist" cannot be asserted directly.
- **Cross-component check with helm-charts:** the chart-rendered `capabilities.json`
  (now carrying the new `remediation` key) parses cleanly through the operator's
  real `LoadCapabilitiesConfig`; unknown keys are tolerated (viper default), exactly
  as the existing `riskAcceptance` key already is.

## Phasing

This is Phase 1 of the merged design. Next: the `kubescape` CLI
`operator remediate annotate` subcommand (completes the user-facing loop), then
Phase 2 `quarantine` + findings-driven targeting, then Phase 3 `cordon` +
operator-native auto-remediation. The `Remediator` registry is designed so each
adds an implementation without touching the command pipeline.

## Summary by CodeRabbit

* **New Features**
  * Added Phase‑1 operator action support for **Annotate** and **Revert** on workloads, including server-side dry-run support and requiring an explicit target.
  * Added action result recording with best-effort event emission for visibility.
  * Added safety validations: excluded namespaces blocked, namespaced kinds must include a namespace, selector targeting not supported, and TTL values are rejected as unsupported.
* **Bug Fixes**
  * Improved request routing so operator actions are handled as a single action request.
* **Tests**
  * Expanded coverage for dry-run vs confirm behavior, revert annotation removal, and input/error cases.
* **Chores**
  * Updated a pinned dependency version.


mainhandler/remediators/annotate.go
```diff
@@ -0,0 +1,157 @@
...
+	// A JSON merge patch deletes a key by setting it to null.
+	patch, err := annotationPatch(map[string]interface{}{
+		AnnotationRemediated: nil,
+		AnnotationReason:     nil,
+		AnnotationFindingRef: nil,
+	})
+	if err != nil {
+		return Result{}, err
+	}
+	if err := r.patch(ctx, t, patch, false); err != nil {
```
matthyx: **Blocker: `Revert` performs a real cluster write regardless of the dry-run flag, breaking the safe-by-default contract.**

`handleOperatorAction` computes `dryRun := args.IsDryRun()` but never passes it down the revert path — `Revert` hard-codes `r.patch(ctx, t, patch, false)`. The doc comment on `handleOperatorAction` states the invariant explicitly: *"a missing flag can never silently perform a real cluster write."* That holds for `annotate` but not for `revert`: a caller that sends a `revert` action without `dryRun=false` (i.e. the default, where `IsDryRun()` returns `true`) expects a preview but gets a real mutation.

Blast radius is low (it only deletes the three `kubescape.io/*` annotations the operator added), but the contract violation is real and the path is reachable today by any command sender, even though the CLI subcommand ships later.

Suggested fix — honor the flag for revert too, e.g. give `Revert` a `dryRun bool` parameter and pass `dryRun` from the switch:
```go
case apis.OperatorActionRevert:
    result, err := r.Revert(ctx, target, dryRun)
```
and inside `Revert`: `r.patch(ctx, t, patch, dryRun)`. If revert is intentionally always-write for Phase 1, please drop the blanket safety claim from the `handleOperatorAction` doc comment and note the exception, so the contract and the code agree.

mainhandler/actionhandler.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	target := remediators.Target{
+		Kind:      args.Target.Kind,
+		Namespace: args.Target.Namespace,
+		Name:      args.Target.Name,
+	}
+
+	// Safety rail: never act on excluded / protected namespaces.
+	if target.Namespace != "" && actionHandler.config.SkipNamespace(target.Namespace) {
```
matthyx, line 70: **Minor: the namespace safety rail is skipped when `target.Namespace == ""`.**

The guard only runs `SkipNamespace` when a namespace is present, while `validateTarget` requires only `Kind` and `Name`. A target with an empty namespace therefore bypasses the excluded-namespace check. In practice the four supported kinds (Deployment/StatefulSet/DaemonSet/Pod) are all namespaced, so the subsequent typed patch with namespace `""` would fail at the API server rather than touch a protected namespace — so this is self-limiting today, not exploitable. Still worth tightening: require a non-empty namespace for namespaced kinds (reject up front), so the rail is enforced rather than relying on an API-server 404.

matthyx (commented):
Reviewed Phase-1 remediation. Pulled the branch: `go build ./...`, `go vet ./...` and the new `mainhandler` / `mainhandler/remediators` tests all pass. The design is clean — the `Remediator` registry, two-layer server-side dry-run, the TTL/selector/quarantine/cordon fences, and the namespace rail are all solid, and the test coverage is good.

One issue rises to blocker level (the `revert` path silently breaks the PR's own safe-by-default contract), plus a couple of minor notes below. Everything else looks good to merge once the revert concern is addressed or explicitly scoped out.

matthyx (commented):
Re-reviewed after `8525ec6` ("fix(remediation): make revert respect dry-run and enforce namespace rail"). Both findings from my previous review are resolved:

- **Revert now honors dry-run.** `Revert` takes a `dryRun` param threaded from `handleOperatorAction`, the interface and `Result` (`DryRun`/`Applied`) are updated accordingly, and a default (no `--confirm`) revert now issues a server-side dry-run instead of a real write — the safe-by-default contract holds for revert too. Covered by `TestHandleOperatorAction_RevertDefaultsToDryRun` and `TestAnnotateRevertDryRun`.
- **Namespace rail enforced up front.** New `IsNamespacedKind` check rejects a namespaced target with no namespace before the excluded-namespace check, so nothing slips past the rail to fail late at the API server. Covered by `TestHandleOperatorAction_NamespacedKindRequiresNamespace`.

Pulled the branch: `go build ./...`, `go vet ./...`, and the `mainhandler` / `mainhandler/remediators` tests all pass. No remaining blockers from me — LGTM for Phase 1.

matthyx (approved):

PR #388, opened 2026-06-28
feat(remediation): add quarantine action (deny-all NetworkPolicy + dry-run)

yugal07:
**Contract:** [armosec/armoapi-go#655 — `operatorAction` command contract](https://github.com/armosec/armoapi-go/pull/655) (merged, `v0.0.720`)
**Builds on:** [kubescape/operator#383 — Phase 1 (annotate + dry-run)](https://github.com/kubescape/operator/pull/383) (merged) · RBAC from [kubescape/helm-charts#855](https://github.com/kubescape/helm-charts/pull/855) (merged)

## Overview

This is **Phase 2** of the merged design: it teaches the operator to execute the
`quarantine` action. Phase 1 ([#383](https://github.com/kubescape/operator/pull/383))
shipped `annotate` and the extensible `Remediator` framework; this PR plugs the
next action into that framework with **no pipeline or transport changes** — a new
`Remediator` added to the registry, exactly as Phase 1 was designed to allow.

`quarantine` isolates a workload by creating a **deny-all `NetworkPolicy`** that
selects the workload's pods (both ingress and egress denied). Per the design's
[resolved open question #3](https://github.com/kubescape/designs-and-proposals/pull/5),
the default is **non-destructive**: the pods are **not** mutated or recreated, so
container state (memory, process trees) is preserved for forensic investigation.
Scale-to-zero / container-stop remains a future, explicit opt-in.

## What changed

- **`mainhandler/remediators/quarantine.go`** *(new)* — `QuarantineRemediator`
  implementing `Plan`/`Apply`/`Revert`:
  - **`Plan`** reads the target's pod selector from the **live** object
    (`Deployment`/`StatefulSet`/`DaemonSet` `spec.selector.matchLabels`, or a
    `Pod`'s own labels) and computes a deny-all `NetworkPolicy` without touching
    the cluster. Reading the live selector isolates the *existing* pods without
    recreating them.
  - **`Apply`** creates the `NetworkPolicy` (deterministic name
    `kubescape-quarantine-<target>`); server-side dry-run when `dryRun=true`;
    idempotent (`AlreadyExists` → success, the workload is already isolated).
  - **`Revert`** deletes that `NetworkPolicy`; a missing policy is a clean no-op
    success; honors server-side dry-run.
  - Audit context (`reason`, `findingRef`, the target) is recorded on the
    `NetworkPolicy`'s labels/annotations.
- **`mainhandler/remediators/remediator.go`** — register `quarantine` in
  `NewRegistry` (one line; `cordon` remains the only stubbed action).
- **`mainhandler/actionhandler.go`** — dispatch `quarantine` through a shared
  `applyRemediation` (Plan → Apply → record), and **extend `revert`** to undo
  *every* reversible action on the target (annotate + quarantine), idempotently.
  This is required because the contract's `revert` verb does not name which action
  it undoes; each `Revert` is a safe no-op when there is nothing to undo, and a
  target object that no longer exists (`NotFound`) is skipped so a leftover
  `NetworkPolicy` is still cleaned up.
- **Tests** — `mainhandler/remediators/quarantine_test.go` (11) and additions to
  `mainhandler/actionhandler_test.go` (5); the existing "unimplemented actions"
  test is narrowed to `cordon` only.

## Safe-by-default behaviour (unchanged from Phase 1)

- **Dry-run is the default.** A command without an explicit `dryRun=false` (the
  CLI's `--confirm`) creates the `NetworkPolicy` with `metav1.DryRunAll` — validated
  against admission, never persisted.
- **Namespace safety rail.** Targets in excluded/protected namespaces
  (`config.SkipNamespace`) are rejected before any client call.
- **No default-install power.** Creating `NetworkPolicy` requires the opt-in
  mutating RBAC from [helm-charts#855](https://github.com/kubescape/helm-charts/pull/855)
  (`capabilities.remediation=enable`), which the default install does not grant —
  so a mistakenly-triggered quarantine fails closed with `Forbidden`, mutating
  nothing. **No new RBAC is needed**: #855 already provisioned
  `networking.k8s.io/networkpolicies` `create/delete/get/list`.
- **No new contract.** `armoapi-go v0.0.720` already carries the `quarantine` verb;
  this PR needs no `armoapi-go` change.

## Additional Information

- **Why deny-all NetworkPolicy and not scale-to-zero.** The design resolved this:
  scaling to zero destroys the container and wipes forensic evidence. A deny-all
  policy cuts the workload off the network while leaving it running for
  investigation. Destructive modes are a later, explicit opt-in.
- **Selector source.** The `NetworkPolicy.spec.podSelector` is the workload's own
  pod-selector labels, read live at `Plan` time. A workload with no selector labels
  is rejected with a clear error rather than producing a policy that would match
  every pod in the namespace.
- **Idempotency & revert symmetry.** Re-quarantining an already-isolated workload
  succeeds; `revert` is safe to call without knowing what was applied and cleans up
  even if the original workload was deleted out from under it.
- **Scope.** Explicit-target `quarantine` only. Findings-driven selector targeting
  (`--control` / `--min-severity`, reading the stored scan-result CRDs) and the
  multi-result status aggregation it requires are the focused next operator PR;
  `cordon` and operator-native auto-remediation are Phase 3.

## How to Test

```bash
go build ./...                       # clean
go vet ./mainhandler/...             # clean
go test ./mainhandler/...            # pass (remediators + mainhandler)
```

Manual end-to-end (cluster with the operator installed and
`capabilities.remediation=enable`), via the CLI once the paired `kubescape`
subcommand lands, or by writing an `OperatorCommand` CRD:

```bash
# after a confirmed quarantine of Deployment payments/api:
kubectl get networkpolicy -n payments kubescape-quarantine-api -o yaml
kubectl get events -n payments --field-selector reason=KubescapeRemediation
```

The `NetworkPolicy` should select the deployment's pod labels with
`policyTypes: [Ingress, Egress]` and no allow rules.

## Related issues/PRs

- Implements (in part): #1770 — CLI control over cluster operations (Phase 2).
- Design: [kubescape/designs-and-proposals#5](https://github.com/kubescape/designs-and-proposals/pull/5) (merged).
- Contract: [armosec/armoapi-go#655](https://github.com/armosec/armoapi-go/pull/655) (merged, `v0.0.720`).
- Builds on: [kubescape/operator#383](https://github.com/kubescape/operator/pull/383) (merged) — Phase 1 framework.
- RBAC/install: [kubescape/helm-charts#855](https://github.com/kubescape/helm-charts/pull/855) (merged) — already grants the NetworkPolicy RBAC.

## Summary by CodeRabbit

* **New Features**
  * Added a new quarantine action that isolates workloads by creating a deny-all network policy.
  * Revert now attempts to undo both annotation and quarantine actions when available.
* **Bug Fixes**
  * Improved revert behavior to handle missing resources gracefully.
  * Quarantine actions now support dry-run and confirm modes consistently.
* **Tests**
  * Added coverage for quarantine planning, apply/revert flows, excluded namespaces, and revert behavior.

### DIFF
--- a/mainhandler/remediators/quarantine.go
+++ b/mainhandler/remediators/quarantine.go
@@ -1,0 +1,213 @@
+package remediators
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/armosec/armoapi-go/apis"
+	networkingv1 "k8s.io/api/networking/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// LabelQuarantine marks the deny-all NetworkPolicy created by the quarantine
+	// action, so it can be found and removed on revert.
+	LabelQuarantine = "kubescape.io/quarantine"
+	// AnnotationQuarantineTarget records, on the NetworkPolicy, the workload it
+	// isolates — part of the audit trail.
+	AnnotationQuarantineTarget = "kubescape.io/quarantine-target"
+
+	// quarantineNPPrefix is the deterministic name prefix of the deny-all
+	// NetworkPolicy, so revert can locate it from the target name alone.
+	quarantineNPPrefix = "kubescape-quarantine-"
+	// maxNameLen is the Kubernetes object-name length limit.
+	maxNameLen = 253
+)
+
+// QuarantineRemediator isolates a workload by creating a deny-all NetworkPolicy
+// that selects the workload's pods (both ingress and egress denied). It does
+// not mutate or recreate the pods, so container state is preserved for forensic
+// investigation (the design's resolved default; scale-to-zero is a future,
+// explicit opt-in). Revert deletes the NetworkPolicy.
+type QuarantineRemediator struct {
+	client kubernetes.Interface
+}
+
+// NewQuarantineRemediator returns a quarantine remediator backed by client.
+func NewQuarantineRemediator(client kubernetes.Interface) *QuarantineRemediator {
+	return &QuarantineRemediator{client: client}
+}
+
+// Plan reads the target workload's pod selector from the live object and
+// computes the deny-all NetworkPolicy without creating it.
+func (r *QuarantineRemediator) Plan(ctx context.Context, req Request) (Plan, error) {
+	if err := validateTarget(req.Target); err != nil {
+		return Plan{}, err
+	}
+	selector, err := r.resolvePodSelector(ctx, req.Target)
+	if err != nil {
+		return Plan{}, err
+	}
+	np := r.buildNetworkPolicy(req, selector)
+	body, err := json.Marshal(np)
+	if err != nil {
+		return Plan{}, err
+	}
+	return Plan{
+		Action:      string(apis.OperatorActionQuarantine),
+		Target:      req.Target,
+		Description: fmt.Sprintf("deny-all NetworkPolicy %q isolating %s (podSelector %v)", np.Name, req.Target, selector),
+		Patch:       string(body),
+	}, nil
+}
+
+// Apply creates the planned deny-all NetworkPolicy. With dryRun=true it is sent
+// as a server-side dry-run (validated against admission, never persisted); only
+// dryRun=false performs a real write. An existing policy is treated as success
+// (the workload is already isolated by a prior quarantine).
+func (r *QuarantineRemediator) Apply(ctx context.Context, p Plan, dryRun bool) (Result, error) {
+	var np networkingv1.NetworkPolicy
+	if err := json.Unmarshal([]byte(p.Patch), &np); err != nil {
+		return Result{}, fmt.Errorf("quarantine: failed to decode planned NetworkPolicy: %w", err)
+	}
+	opts := metav1.CreateOptions{}
+	if dryRun {
+		opts.DryRun = []string{metav1.DryRunAll}
+	}
+	_, err := r.client.NetworkingV1().NetworkPolicies(np.Namespace).Create(ctx, &np, opts)
+	switch {
+	case err == nil, apierrors.IsAlreadyExists(err):
+		// created, or already isolated by a prior quarantine — both are success.
+	default:
+		return Result{}, fmt.Errorf("quarantine: failed to create NetworkPolicy %q: %w", np.Name, err)
+	}
+	return Result{
+		Action:      p.Action,
+		Target:      p.Target,
+		DryRun:      dryRun,
+		Applied:     !dryRun,
+		Description: p.Description,
+	}, nil
+}
+
+// Revert deletes the deny-all NetworkPolicy that quarantined the target. A
+// missing policy is treated as success (nothing to undo). Like Apply, dryRun=true
+// issues a server-side dry-run delete, so the safe-by-default contract holds for
+// revert too.
+func (r *QuarantineRemediator) Revert(ctx context.Context, t Target, dryRun bool) (Result, error) {
+	if err := validateTarget(t); err != nil {
+		return Result{}, err
+	}
+	name := quarantineNPName(t.Name)
+	opts := metav1.DeleteOptions{}
+	if dryRun {
+		opts.DryRun = []string{metav1.DryRunAll}
+	}
+	err := r.client.NetworkingV1().NetworkPolicies(t.Namespace).Delete(ctx, name, opts)
+	applied := !dryRun
+	desc := fmt.Sprintf("deleted quarantine NetworkPolicy %q in namespace %q", name, t.Namespace)
+	switch {
+	case apierrors.IsNotFound(err):
+		applied = false
+		desc = fmt.Sprintf("no quarantine NetworkPolicy %q in namespace %q; nothing to revert", name, t.Namespace)
+	case err != nil:
+		return Result{}, fmt.Errorf("quarantine: failed to delete NetworkPolicy %q: %w", name, err)
+	}
+	return Result{
+		Action:      string(apis.OperatorActionRevert),
+		Target:      t,
+		DryRun:      dryRun,
+		Applied:     applied,
+		Description: desc,
+	}, nil
+}
+
+// resolvePodSelector returns the labels that select the target's running pods,
+// read from the live object so quarantine isolates the existing pods without
+// recreating them.
+func (r *QuarantineRemediator) resolvePodSelector(ctx context.Context, t Target) (map[string]string, error) {
+	var labels map[string]string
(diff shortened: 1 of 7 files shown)


mainhandler/remediators/quarantine.go
```diff
@@ -0,0 +1,215 @@
...
+		if err != nil {
+			return nil, err
+		}
+		selector = o.Spec.Selector
+	case "pod":
+		o, err := r.client.CoreV1().Pods(t.Namespace).Get(ctx, t.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+		selector = &metav1.LabelSelector{MatchLabels: o.Labels}
```
matthyx, line 232: **Pod target isolates the whole ReplicaSet, not the single pod.** For a Pod target the `podSelector` is the pod's *full* label set, which for a controller-managed pod includes `pod-template-hash` (or the StatefulSet/DaemonSet controller labels). That selects every replica sharing the template, so `quarantine Pod payments/api-0` actually cuts off *all* of api's pods — `TestQuarantineApplyPodUsesPodLabels` encodes exactly this (`{app: api, pod-template-hash: abc}`). NetworkPolicy can't select an individual pod by name, so single-pod forensic isolation isn't achievable this way. If isolating one pod while keeping the rest of the service up is a goal, this doesn't meet it; if it isn't, the Pod kind should at least document that it isolates the entire template. Worth a decision before merge.

mainhandler/remediators/quarantine.go
```diff
@@ -0,0 +1,215 @@
...
+	if err := json.Unmarshal([]byte(p.Patch), &np); err != nil {
+		return Result{}, fmt.Errorf("quarantine: failed to decode planned NetworkPolicy: %w", err)
+	}
+	opts := metav1.CreateOptions{}
+	if dryRun {
+		opts.DryRun = []string{metav1.DryRunAll}
+	}
+	_, err := r.client.NetworkingV1().NetworkPolicies(np.Namespace).Create(ctx, &np, opts)
+	switch {
+	case err == nil, apierrors.IsAlreadyExists(err):
```
matthyx: **`AlreadyExists` masks selector drift.** If the workload's `spec.selector` changed since the first quarantine (edited/redeployed), `Plan` builds an NP with the new selector but this branch short-circuits to success, leaving the **old** NetworkPolicy — with the stale selector — in place. The currently-running pods are then not actually isolated even though the action returns `Applied: true`. Consider reconciling the existing policy (update/replace, or compare specs) instead of treating any pre-existing name as an unconditional success.

matthyx (commented):
Reviewed Phase 2 (quarantine). The structure is clean — the Remediator framework absorbs the new action with no pipeline change, safe-by-default dry-run is preserved, revert is idempotent and NotFound-tolerant, deterministic naming disambiguates by kind and hash-truncates long names, and `matchExpressions` are carried through. Test coverage is good.

I verified the RBAC claim end-to-end: `Plan` reads the live workload via `Get`, which is **not** covered by the remediation role in helm-charts#855 (that grants only `patch` on workloads + create/delete on networkpolicies) — but the operator's **base** ClusterRole already grants `get/watch/list` on deployments/statefulsets/daemonsets/pods, so quarantine's read is covered. RBAC is fine; no change needed there.

That said, a few isolation-semantics issues are worth resolving before this ships as an incident-response control:

**1. Selector overlap → collateral isolation (blast radius).** The deny-all NetworkPolicy applies to *every* pod in the namespace matching the workload's selector labels, not only the target workload's pods. A Deployment's `spec.selector` is typically a strict subset of the pod labels (e.g. `{app: api}`), so any other workload sharing those labels (blue/green, canary, a second Deployment reusing `app=`) gets isolated too. For a "quarantine this one workload" action that's unbounded collateral. Worth a guard (e.g. reject/verify when the selector also matches pods the target doesn't own) or an explicit, documented assumption.

**2. CNI enforcement dependency → false assurance.** A deny-all NetworkPolicy is a silent no-op on clusters whose CNI doesn't enforce NetworkPolicy (plain flannel, some managed offerings). The action still creates the policy and reports `Applied: true`, so an operator believes a compromised workload is isolated when it isn't. For a security control this is the most dangerous failure mode. Recommend documenting the CNI precondition and, ideally, surfacing it in the result.

**3. Minor (audit text):** `revert` always runs `annotate.Revert`, whose Result reads "removed kubescape remediation annotations from <target>" with `Applied: true` even for a workload that was only quarantined and never annotated. The emitted Event/status text is then slightly misleading. Consider only reporting sub-actions that actually changed something.

See the two inline notes for the Pod-target and re-quarantine specifics.

matthyx (approved):
Re-reviewed after `e61a4dd` — all five points from my earlier review are addressed:

1. **Selector-overlap blast radius** — now documented in the `QuarantineRemediator` type doc and surfaced at runtime via `quarantineCaveat` in the plan/result Description, so the audit trail states the assumption. Reasonable resolution (a hard uniqueness guard would be nice-to-have, but making the assumption explicit is a fair call for an explicit-target action).
2. **Pod target isolates the whole ReplicaSet** — documented in the type doc and inline at the pod case; the limitation is now explicit rather than surprising.
3. **CNI enforcement / false assurance** — documented and surfaced in the caveat.
4. **`AlreadyExists` masking selector drift** — fixed for real: `Apply` now calls `reconcile()`, which Gets the existing policy, `DeepEqual`-compares spec/labels/annotations, and Updates on drift (honoring dry-run); a policy that vanished mid-flight is a clean no-op. Covered by `TestQuarantineApplyReconcilesSelectorDrift` and `TestQuarantineApplyReconcileHonorsDryRun`.
5. **Misleading revert audit** — `annotate.Revert` now reads the live annotations first and reports `Applied: false` / "nothing to revert" when none are present, and returns `NotFound` up to the handler so leftover-NP cleanup still works. Covered by `TestAnnotateRevertNoAnnotationsIsNoop`.

integration-build (which runs the Go suite) is green on the latest commit. No new concerns. LGTM.

matthyx (approved):

matthyx:
@yugal07 do you need a release for this?

yugal07:
Yes we'd have to consume this in kubescape

yugal07:
I think the chain would be: once this is merged, the operator image is built automatically — pr-merged.yaml pushes quay.io/kubescape/operator:v0.2.<run-number> on merge to main, so there's no separate tag/release step. From there it's a helm-charts bump of operator.image.tag to that new tag; no RBAC change is needed since #855 already provisions networkpolicies create/delete/get/list under capabilities.remediation: enable (and the base role already covers the workload reads). Once that chart is out and deployable, the paired kubescape CLI quarantine subcommand can be integration-tested end-to-end and shipped it only needs the armoapi-go contract, already at v0.0.720. So: merge → helm-charts tag bump → wire up + e2e the CLI side. I'll take the CLI side once the merge lands and the chart points at the new image.

PR #391, opened 2026-07-10
operator: findings-driven targeting for `operatorAction` (resolve `selector` from stored scan results)

yugal07:
**Contract:** [armosec/armoapi-go#655 — `operatorAction` command contract](https://github.com/armosec/armoapi-go/pull/655) (merged, `v0.0.720`)
**Builds on:**
- [kubescape/operator#383 — Phase 1 remediation (annotate + `Remediator` framework)](https://github.com/kubescape/operator/pull/383) (merged)
- [kubescape/operator#388 — `quarantine` action (deny-all NetworkPolicy)](https://github.com/kubescape/operator/pull/388) (merged)
- [kubescape/helm-charts#855 — opt-in `capabilities.remediation` + gated RBAC](https://github.com/kubescape/helm-charts/pull/855) (merged)

## Overview

This completes the second half of **Phase 2** of the merged design: **findings-driven
targeting**. The operator can already execute `annotate`/`quarantine`/`revert`
against an *explicit* target (`{kind, namespace, name}`). This PR teaches it to
resolve the design's findings-driven **`selector`** — "act on every workload that
fails control `C-0016`" or "…that has a `High`-or-worse finding" — into a concrete
target set, then run the requested action against each.

No re-scan and no new transport: the selector is resolved by reading the scan
results **already stored in the cluster** — the `workloadconfigurationscansummaries`
CRDs in `spdx.softwarecomposition.kubescape.io` — which the operator already has
read access to. Remediation stays a verb on the existing `apis.Command` pipeline.

Until now the handler explicitly rejected the selector:

```go
// before
return fmt.Errorf("operatorAction: findings-driven selector targeting is not supported yet (phase 2); provide an explicit target")
```

With this change, a command carrying a selector instead of a target is executed:

```jsonc
// Command.Args for "annotate every workload failing C-0016 at High or above"
{
  "action":   "annotate",
  "selector": { "control": "C-0016", "minSeverity": "High" },
  "reason":   "C-0016 allowPrivilegeEscalation",
  "dryRun":   true                     // default; only an explicit false writes
}
```

The operator resolves the selector against stored summaries, then applies
`annotate` (dry-run by default) to each matching workload.

## How selection works

A configuration-scan summary matches the selector when:

| Selector | Match rule |
|---|---|
| `control` only | the named control is present **and failing** on that workload |
| `minSeverity` only | the workload has **any failing finding** at or above that severity |
| both | the named control is failing **and** its severity is at or above `minSeverity` |

- **Severity order:** `Critical > High > Medium > Low > Unknown`, so `minSeverity`
  is a simple `>=` floor (`High` matches `Critical` and `High`, excludes `Medium`).
- **Empty selector is rejected, never "match everything".** A selector with
  neither `control` nor `minSeverity` returns an error rather than resolving to
  every workload — the single most important safety guard here.
- **Selector that matches nothing is an explicit error**, not a silent no-op, so
  the caller learns the action had no effect.
- Each resolved workload is mapped back from the summary's
  `kubescape.io/workload-{kind,name,namespace}` labels; summaries missing those
  labels, non-remediable kinds (e.g. `CronJob`), and duplicates (re-scans of the
  same workload) are dropped during resolution so they never reach a remediator.

## What changed

- **`mainhandler/findings.go`** *(new)* — `resolveSelectorTargets`: lists
  `WorkloadConfigurationScanSummaries` cluster-wide, filters by control/severity
  (`summaryMatchesSelector`, `findFailedControl`, `summaryHasFailingSeverityAtLeast`),
  maps matches to `remediators.Target`, and drops excluded namespaces /
  non-remediable kinds / duplicates. Plus the `severityRank` helper.
- **`mainhandler/actionhandler.go`** — `handleOperatorAction` now resolves the
  target set via a new `resolveTargets` (explicit `Target` **or** `Selector`) and
  dispatches each through a new `handleActionOnTarget` (the per-target safety
  rails + action switch, factored out of the old single-target body). A single
  target returns its error verbatim (existing behavior byte-for-byte); a
  multi-target selector runs every target and `errors.Join`s any failures so one
  bad workload doesn't abandon the rest.
- **`mainhandler/handlerequests.go`** — thread the existing `ksStorageClient`
  (`kssc.Interface`) into `ActionHandler` (struct field + `NewActionHandler`
  parameter + its one call site) so the handler can read the scan summaries.
- **Tests** — `findings_test.go` (new) unit-tests the resolver and severity
  ranking; `actionhandler_test.go` replaces the obsolete "not supported yet"
  assertion with end-to-end selector cases (annotates only the failing workload,
  leaves passers untouched, defaults to dry-run, errors when nothing matches).

## Safe-by-default (unchanged)

- **Dry-run is still the default** for selector actions: with `dryRun` unset,
  every matched workload is patched with server-side dry-run
  (`metav1.DryRunAll`), never a real write. Only an explicit `dryRun=false` (the
  CLI's `--confirm`) writes.
- **Namespace rails apply per target.** Excluded/protected namespaces are dropped
  during resolution *and* re-checked before any write, so a selector can never
  reach a protected namespace.
- **RBAC is unchanged.** The read the resolver needs (`list` on
  `workloadconfigurationscansummaries`) is already in the operator's **base**
  ClusterRole — no `helm-charts` change is required, and the mutating verbs stay
  gated behind `capabilities.remediation` from
  [helm-charts#855](https://github.com/kubescape/helm-charts/pull/855). With
  remediation disabled the operator can read findings but still cannot mutate, so
  a selector action fails closed with `Forbidden`.

## Not in scope (follow-ups)

- **CLI flags (`--control` / `--min-severity`).** This is the operator half; the
  paired `kubescape` PR exposes the selector on the `operator remediate`
  subcommand. The operator now understands the selector on the wire regardless of
  which client sends it.
- **Vulnerability-summary targeting.** Selection reads posture
  (`workloadconfigurationscansummaries`), which is what `control` and
  `minSeverity` map to. Driving `minSeverity` off `vulnerabilitymanifestsummaries`
  is a natural extension of the same resolver.
- **`cordon`** — still stubbed; a node action is not workload-findings-driven and
  is left for its own phase.
- **Aggregated multi-target status.** Each resolved workload writes its own
  Kubernetes Event; the single `OperatorCommand` status payload currently reflects
  the last target (a partial failure still surfaces via the joined error). A
  structured per-target status is a reasonable later refinement.

## Testing

- `go build ./...`, `go vet ./mainhandler/...`, and `gofmt` are clean;
  `go mod tidy` is a no-op (all new imports were already direct dependencies).
- `go test -race -count=1 ./mainhandler/ ./mainhandler/remediators/` passes.
- The full `go test ./...` passes for every package except
  `watcher/TestRegistryCommandWatch`, a pre-existing k3s/testcontainers
  integration test that times out booting its container in this environment
  (fails at `k3s.Run(...)` before any operator logic; `watcher` does not import
  `mainhandler`, so it is unaffected by this change).
- New tests cover: control-only, severity-only, and combined selection; the
  empty-selector guard; invalid `minSeverity`; excluded-namespace and
  non-remediable-kind skipping; label-less summaries; de-duplication; and the
  end-to-end handler path (only failing workloads annotated, dry-run default,
  match-nothing error).

## Summary by CodeRabbit

- **New Features**
  - Added selector-based actions for targeting workloads based on stored configuration findings.
  - Supports filtering by control and minimum severity.
  - Actions can now run across multiple matching workloads, with individual failures reported together.
  - Automatically skips excluded namespaces, unsupported workload types, and incomplete finding records.
  - Selector actions default to server-side dry-run when no dry-run setting is provided.

- **Bug Fixes**
  - Reports clear errors when selectors are empty or match no workloads.
  - Prevents duplicate actions against the same workload.

### DIFF
--- a/mainhandler/actionhandler.go
+++ b/mainhandler/actionhandler.go
@@ -3,6 +3,7 @@ package mainhandler
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -46,22 +47,61 @@ func (actionHandler *ActionHandler) handleOperatorAction(ctx context.Context) er
 		return fmt.Errorf("operatorAction: ttl/auto-revert is not supported yet (later phase); omit ttl")
 	}
 
-	// Phase 1 ships explicit-target actions only. Findings-driven Selector
-	// targeting (reading the stored scan-result CRDs) arrives in phase 2.
-	if args.Target == nil {
-		if args.Selector != nil {
-			return fmt.Errorf("operatorAction: findings-driven selector targeting is not supported yet (phase 2); provide an explicit target")
+	// Resolve the target set: an explicit Target, or a findings-driven Selector
+	// that reads the stored configuration-scan summaries (no re-scan). Exactly
+	// one of the two must be provided.
+	targets, err := actionHandler.resolveTargets(ctx, args)
+	if err != nil {
+		return err
+	}
+
+	dryRun := args.IsDryRun()
+	registry := remediators.NewRegistry(actionHandler.k8sAPI.KubernetesClient)
+
+	// A single target (explicit, or a selector that resolved to one) returns its
+	// error verbatim. A multi-target selector runs every target and aggregates
+	// failures so one bad workload does not abandon the rest.
+	if len(targets) == 1 {
+		return actionHandler.handleActionOnTarget(ctx, args, targets[0], registry, dryRun)
+	}
+	var errs []error
+	for _, target := range targets {
+		if err := actionHandler.handleActionOnTarget(ctx, args, target, registry, dryRun); err != nil {
+			errs = append(errs, fmt.Errorf("target %s: %w", target, err))
 		}
-		return fmt.Errorf("operatorAction: a target is required")
 	}
+	return errors.Join(errs...)
+}
 
-	target := remediators.Target{
-		Kind:      args.Target.Kind,
-		Namespace: args.Target.Namespace,
-		Name:      args.Target.Name,
+// resolveTargets returns the concrete objects the action operates on. An
+// explicit Target is used as-is; a Selector is resolved against stored findings.
+// A selector that matches nothing is an error, not a silent no-op, so the caller
+// learns the action had no effect.
+func (actionHandler *ActionHandler) resolveTargets(ctx context.Context, args apis.OperatorActionArgs) ([]remediators.Target, error) {
+	if args.Target != nil {
+		return []remediators.Target{{
+			Kind:      args.Target.Kind,
+			Namespace: args.Target.Namespace,
+			Name:      args.Target.Name,
+		}}, nil
 	}
+	if args.Selector != nil {
+		targets, err := actionHandler.resolveSelectorTargets(ctx, args.Selector)
+		if err != nil {
+			return nil, err
+		}
+		if len(targets) == 0 {
+			return nil, fmt.Errorf("operatorAction: selector matched no workloads (control=%q minSeverity=%q)", args.Selector.Control, args.Selector.MinSeverity)
+		}
+		return targets, nil
+	}
+	return nil, fmt.Errorf("operatorAction: a target is required")
+}
 
-	// All Phase-1 target kinds are namespaced. Require the namespace up front so
+// handleActionOnTarget enforces the per-target safety rails and dispatches the
+// action to the matching remediator. It runs once per resolved target.
+func (actionHandler *ActionHandler) handleActionOnTarget(ctx context.Context, args apis.OperatorActionArgs, target remediators.Target, registry map[apis.OperatorActionType]remediators.Remediator, dryRun bool) error {
+	// All current target kinds are namespaced. Require the namespace up front so
 	// the excluded-namespace rail below is actually enforced, instead of an empty
 	// namespace slipping past it and failing late at the API server.
 	if remediators.IsNamespacedKind(target.Kind) && target.Namespace == "" {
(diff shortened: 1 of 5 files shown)


matthyx (commented):
## Review — findings-driven `selector` targeting

I reviewed this end-to-end (built, `go vet`, `gofmt`, and `go test ./mainhandler/...` all clean on `pr391`). It's well-structured and safe-by-default; the explicit-target path is preserved byte-for-byte and the selector path is nicely guarded (empty-selector rejected, invalid `minSeverity` rejected, match-nothing is an explicit error, excluded namespaces / non-remediable kinds / label-less summaries / duplicates all dropped before any remediator runs).

**No merge blockers found.** One safety-critical thing to confirm, plus a couple of non-blocking notes.

### Please confirm — `minSeverity`-only relies on `Spec.Severities` counting *failed* controls
`summaryHasFailingSeverityAtLeast` (findings.go) matches a workload when `summary.Spec.Severities.{Critical,High,…} > 0`. This is correct **only if** those counters count *failed* controls per severity — which is the kubescape convention, so I believe it's right. But it's the one assumption that, if wrong, would target clean workloads: a summary that merely *passed* a High control would falsely match `minSeverity: High`.

Note the unit test (`TestResolveSelectorTargets_ByMinSeverity`) can't catch a mismatch here — it hand-builds the `Severities` block with the same assumption, so it validates the resolver, not the scanner's real output. Worth an explicit confirmation against how the config-scan summarizer populates `Severities`. Dry-run being the default mitigates the blast radius either way. The control path (`findFailedControl` → status `failed`) is unambiguous and doesn't have this dependency.

### Non-blocking

- **`Target` + `Selector` both set → `Selector` silently ignored.** `resolveTargets` returns the explicit target without checking `Selector == nil`, so a caller sending both gets single-target behavior with no error. Consider rejecting the ambiguous combination (or documenting the precedence) rather than silently dropping the selector.

- **Unbounded cluster-wide `List`** (already flagged by CodeRabbit). Fine for this phase; worth tracking a `Limit`/continue-token pager as the feature scales to large clusters.

- **Partial-failure observability** — you already call this out as a follow-up: on a multi-target selector, failed targets emit no Kubernetes Event and the `OperatorCommand` payload reflects the last *successful* target, though the joined error does flip the command's final status to error. A structured per-target status later would close the gap.

Nice test coverage overall — control-only, severity-only, combined-below-floor, dedup, and the end-to-end dry-run default are all the right cases to pin.

matthyx (approved):

PR #392, opened 2026-07-10
[ LFX 2026 ] operator: watch SecurityException/ClusterSecurityException CRDs and rescan on change or expiry

yugal07:
**Related issue:** [kubescape/kubescape#1982 — Implement SecurityException CRD: kubescape posture path, operator watcher, and Headlamp plugin](https://github.com/kubescape/kubescape/issues/1982)
**Design:** [kubevuln/docs/security-exception-design.md](https://github.com/kubescape/kubevuln/blob/main/docs/security-exception-design.md) — §Architecture → *operator — Watch & Rescan*, §Observability → *Expiry*
**Completes on the scanner side:**
- [kubescape/kubescape#2366 / #2480 — `CRDExceptionsGetter` posture path](https://github.com/kubescape/kubescape/pull/2480) (merged) — reads the same CRDs at scan time
- [kubescape/helm-charts#817 / #859 / #875 — CRDs, CEL validation, RBAC](https://github.com/kubescape/helm-charts/pull/875) (merged) — ships the CRDs this watcher lists/watches

## Overview

This PR adds the **operator half** of the SecurityException design (issue kubescape/kubescape/1982
checklist item *"Operator `SecurityExceptionWatchHandler` watching both kinds with
a debounced `CooldownQueue`, dispatching rescans on changes"*).

The posture path already reads `SecurityException` / `ClusterSecurityException`
CRDs **at scan time** (`CRDExceptionsGetter`). But nothing triggers a scan when an
exception changes: a team commits a new exception via GitOps, and the excepted
findings keep showing as failing until the next *scheduled* scan hours later — or
an exception passes its `expiresAt` and the now-unsuppressed findings stay hidden
until the next schedule. The GitOps loop the design promises ("exceptions live
alongside app code and take effect when applied") was open on the runtime side.

This adds `SecurityExceptionWatchHandler`: it watches both kinds, debounces bursts
of changes, and dispatches a cluster posture rescan so results reflect the current
set of exceptions promptly — plus a periodic expiry sweep so lapsed exceptions
resurface their findings without waiting for a schedule.

It is wired into the operator's existing watcher startup, gated on the posture
scanner being enabled:

```go
// mainhandler/handlerequests.go — HandleWatchers
if mainHandler.config.Components().Kubescape.Enabled {
    seWatch := watcher.NewSecurityExceptionWatchHandler(
        mainHandler.config, mainHandler.k8sAPI.GetDynamicClient(), mainHandler.eventWorkerPool)
    go seWatch.SecurityExceptionWatch(ctx)
}
```

## How it works

```text
securityexceptions ─┐
                    ├─ watch (dynamic client) ─► CooldownQueue ─► rescanSignal ─► TypeRunKubescape
clustersecurity...──┘   + list-existing on start   (per-object   (coalesce+     (cluster WildWlid,
                                                     debounce)     throttle)      allcontrols/nsa/mitre)
expiry sweep (5m) ──────────────────────────────────────────────►┘
```

- **List-existing + watch, both kinds.** On start the handler lists current
  objects (so exceptions present before the operator came up are honored) then
  follows changes on `securityexceptions` and `clustersecurityexceptions` at
  `kubescape.io/v1beta1` via the dynamic client, re-establishing a dropped watch
  with a bounded retry.
- **Debounce (design's `CooldownQueue`).** Every event goes through the operator's
  existing `CooldownQueue`, which collapses repeated events for the *same* object
  during a cooldown window.
- **Coalesce + throttle across objects.** A GitOps apply touches many CRDs at once;
  each is a distinct queue key, so distinct events would otherwise each trigger a
  full rescan. A buffered-by-one `rescanSignal` collapses concurrent requests into a
  single pending rescan, and a post-dispatch throttle absorbs the rest of the burst
  — a set of exceptions applied together produces **one** rescan, not N.
- **Rescan = the operator's own startup scan.** `buildRescanCommand` emits exactly
  the `apis.TypeRunKubescape` command the operator already fires on startup
  (cluster `WildWlid`, frameworks `allcontrols`/`nsa`/`mitre`), dispatched through
  the existing `utils.AddCommandToChannel` worker pool. Re-evaluating the whole
  cluster is idempotent and re-applies every exception; no new transport or command
  type is introduced.
- **Expiry sweep (design's expiry controller).** A 5-minute ticker lists both kinds
  and requests a rescan when an exception has newly passed its `spec.expiresAt`
  (parsed as RFC3339). Consistent with the design, **nothing writes status** —
  expiry is still evaluated by the scanners at scan time; the sweep only nudges a
  rescan so previously-suppressed findings resurface. Each expiry triggers exactly
  one rescan (tracked in `expiredSeen`), and the mark is cleared if `expiresAt` is
  later pushed back out, so a re-expiry is caught again.

## What changed

- **`watcher/securityexceptionwatcher.go`** *(new)* — `SecurityExceptionWatchHandler`:
  the two watches (`watchKind` / `consumeWatch` / `listExisting`), the debounced
  event handler (`handleEvents` → `requestRescan`), the coalescing rescan dispatcher
  (`rescanLoop`), the expiry loop (`expiryLoop` / `sweepExpired`), and the pure
  helpers `buildRescanCommand` and `parseExpiresAt`. Tunables (cooldown, throttle,
  retry, sweep interval) are captured into the handler at construction so a running
  handler is unaffected by later changes and tests are race-free.
- **`mainhandler/handlerequests.go`** — start the handler in `HandleWatchers`, gated
  on `Components().Kubescape.Enabled` (the posture scanner an exception affects). The
  dispatcher is wired to the existing `eventWorkerPool`; `dispatchRescan` is an
  injectable field so tests exercise the pipeline without a pool.
- **`watcher/securityexceptionwatcher_test.go`** *(new)* — unit tests (below).

No new RBAC: the operator already has `get`/`list`/`watch` on the SecurityException
CRDs from [helm-charts#875](https://github.com/kubescape/helm-charts/pull/875), and
dispatches through the same pool it already uses for scans.

## Design decisions

- **Full cluster rescan, not per-workload targeting.** The design mentions
  "determines affected namespaces/workloads, dispatches rescan commands." Resolving
  an exception's `resources`/`objectSelector`/`namespaceSelector` to a concrete
  workload set duplicates matching logic that already lives in the scanner and
  opa-utils. A cluster rescan is simpler, always correct, idempotent, and coalesced
  so it isn't wasteful. Targeted rescans are a clean follow-up (see below).
- **`v1beta1`.** The handler lists/watches at `kubescape.io/v1beta1`, matching the
  shipped CRDs and `CRDExceptionsGetter`. A single named constant moves when the CRD
  graduates to `v1`.
- **Gated on `Kubescape.Enabled`.** An exception only affects posture results, so
  the watcher runs exactly when the posture scanner does — no new config flag.

## Testing

- `go build ./...`, `go vet ./...`, and `gofmt` are clean.
- `go test -race -count=1 ./watcher/` passes for the new suite (9 tests). The one
  package failure, `watcher/TestRegistryCommandWatch`, is a **pre-existing**
  testcontainers integration test that times out booting its container in this
  environment (confirmed failing identically on a clean `main`); it is unrelated to
  this change.
- **Unit tests** (`securityexceptionwatcher_test.go`) cover: `buildRescanCommand`
  shape; `parseExpiresAt` (present/absent/malformed/date-only); rescan coalescing
  (5 requests → 1 pending); the expiry sweep (one-shot per expiry, idempotent on
  re-sweep, and re-arm when `expiresAt` is pushed out); `listExisting` enqueue;
  event→rescan dispatch; and the full `SecurityExceptionWatch` path against a fake
  dynamic client. Two real concurrency issues surfaced by `-race` were fixed during
  development: a shared `expiredSeen` map (now mutex-guarded) and loops reading
  mutable package globals (now captured into the handler).

### End-to-end, against a live kind cluster

Run with the released operator replaced by this build's handler (driven via a small
`go run` harness that constructs `NewSecurityExceptionWatchHandler` with the real
dynamic client and prints each dispatched command):

```text
=== watching SecurityException/ClusterSecurityException for 45s ===
[info] security exception changed, requesting rescan. kind: ClusterSecurityException; name: staging-relax-c0034; type: ADDED
>>> RESCAN DISPATCHED  cmd=kubescapeScan wildWlid=wlid://cluster-kind-kubescape-test/namespace-
[info] security exception changed, requesting rescan. kind: SecurityException; name: watch-probe; namespace: production; type: ADDED
>>> RESCAN DISPATCHED  cmd=kubescapeScan wildWlid=wlid://cluster-kind-kubescape-test/namespace-
[info] security exception changed, requesting rescan. kind: SecurityException; name: watch-probe; namespace: production; type: DELETED
>>> RESCAN DISPATCHED  cmd=kubescapeScan wildWlid=wlid://cluster-kind-kubescape-test/namespace-
```

- **List-existing on start** found the pre-existing `ClusterSecurityException` and
  dispatched a rescan.
- **Apply** and **delete** of a `SecurityException` each produced a debounced rescan
  with the correct cluster `WildWlid`.
- Paired with a `kubescape scan` against the same cluster, the dispatched rescan
  flips an excepted control (e.g. `C-0034` on `Deployment/nginx`) from `failed` to
  `passed (w/exceptions)` and back on delete — closing the GitOps loop end to end.

## Summary by CodeRabbit

* **New Features**
  * Added monitoring for SecurityException and ClusterSecurityException resources (runs when Kubescape and risk acceptance are enabled).
  * Security exception changes now trigger posture rescans.
  * Introduced automatic expiration handling so expired exceptions trigger rescans exactly once per expired state.

* **Bug Fixes**
  * Added debounced/coalesced rescan requests to help prevent rescan storms and re-arm rescans after throttling.

* **Tests**
  * Added a full test suite covering command construction, `expiresAt` parsing, event-driven rescan triggering, startup behavior, and idempotent expiration sweeps.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -29,6 +29,7 @@ type Capabilities struct {
 	Seccomp              string `json:"seccomp"`
 	VulnerabilityScan    string `json:"vulnerabilityScan"`
 	AdmissionController  string `json:"admissionController"`
+	RiskAcceptance       string `json:"riskAcceptance"`
 }
 
 type Components struct {
@@ -150,6 +151,7 @@ type IConfig interface {
 	AdmissionControllerEnabled() bool
 	ContinuousScanEnabled() bool
 	NodeSbomGenerationEnabled() bool
+	RiskAcceptanceEnabled() bool
 	CleanUpRoutineInterval() time.Duration
 	MatchingRulesFilename() string
 	TriggerSecurityFramework() bool
@@ -196,6 +198,15 @@ func (c *OperatorConfig) AdmissionControllerEnabled() bool {
 
 func (c *OperatorConfig) NodeSbomGenerationEnabled() bool {
 	return c.components.Capabilities.NodeSbomGeneration == "enable"
+}
+
+// RiskAcceptanceEnabled reports whether the riskAcceptance capability is on. The
+// SecurityException watcher must gate on this because the operator's RBAC for
+// securityexceptions/clustersecurityexceptions is only granted by the Helm chart
+// when riskAcceptance is enabled — starting the watcher otherwise yields a
+// permission-denied watch loop with no rescans.
+func (c *OperatorConfig) RiskAcceptanceEnabled() bool {
+	return c.components.Capabilities.RiskAcceptance == "enable"
 }
 
 func (c *OperatorConfig) KubevulnURL() string {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -10,6 +10,28 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestRiskAcceptanceEnabled(t *testing.T) {
+	tests := []struct {
+		value string
+		want  bool
+	}{
+		{"enable", true},
+		{"disable", false},
+		{"", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.value, func(t *testing.T) {
+			cfg := NewOperatorConfig(
+				CapabilitiesConfig{Capabilities: Capabilities{RiskAcceptance: tt.value}},
+				armometadata.ClusterConfig{},
+				&utils.Credentials{},
+				Config{},
+			)
+			assert.Equal(t, tt.want, cfg.RiskAcceptanceEnabled())
+		})
+	}
+}
+
 func TestLoadCapabilities(t *testing.T) {
 	type args struct {
 		path string
--- a/mainhandler/handlerequests.go
+++ b/mainhandler/handlerequests.go
@@ -163,6 +163,14 @@ func (mainHandler *MainHandler) HandleWatchers(ctx context.Context) {
 		go watchHandler.ContainerProfileWatch(ctx, mainHandler.eventWorkerPool)
 	}
 
+	// Watch SecurityException/ClusterSecurityException CRDs and rescan posture on
+	// change or expiry, so exception edits take effect without waiting for the
+	// next scheduled scan.
+	if mainHandler.config.Components().Kubescape.Enabled {
+		securityExceptionWatchHandler := watcher.NewSecurityExceptionWatchHandler(mainHandler.config, mainHandler.k8sAPI.GetDynamicClient(), mainHandler.eventWorkerPool)
+		go securityExceptionWatchHandler.SecurityExceptionWatch(ctx)
+	}
+
 	go operatorCommandsHandler.Start()
 	go commandWatchHandler.CommandWatch(ctx)
 }
--- a/mainhandler/handlerequests.go
+++ b/mainhandler/handlerequests.go
@@ -165,8 +165,12 @@ func (mainHandler *MainHandler) HandleWatchers(ctx context.Context) {
 
 	// Watch SecurityException/ClusterSecurityException CRDs and rescan posture on
 	// change or expiry, so exception edits take effect without waiting for the
-	// next scheduled scan.
-	if mainHandler.config.Components().Kubescape.Enabled {
+	// next scheduled scan. Gated on the riskAcceptance capability to match the
+	// Helm chart, which only grants the operator get/list/watch on these CRDs when
+	// riskAcceptance is enabled; starting the watcher without that RBAC would just
+	// loop on permission-denied. Kubescape must also be enabled since the rescan
+	// dispatches a posture scan.
+	if mainHandler.config.RiskAcceptanceEnabled() && mainHandler.config.Components().Kubescape.Enabled {
 		securityExceptionWatchHandler := watcher.NewSecurityExceptionWatchHandler(mainHandler.config, mainHandler.k8sAPI.GetDynamicClient(), mainHandler.eventWorkerPool)
 		go securityExceptionWatchHandler.SecurityExceptionWatch(ctx)
 	}
--- a/watcher/securityexceptionwatcher.go
+++ b/watcher/securityexceptionwatcher.go
@@ -1,0 +1,364 @@
+package watcher
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/armosec/armoapi-go/apis"
+	"github.com/armosec/utils-go/boolutils"
+	pkgwlid "github.com/armosec/utils-k8s-go/wlid"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+	v1 "github.com/kubescape/opa-utils/httpserver/apis/v1"
+	utilsmetav1 "github.com/kubescape/opa-utils/httpserver/meta/v1"
+	"github.com/kubescape/operator/config"
+	"github.com/kubescape/operator/utils"
+	"github.com/panjf2000/ants/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/pager"
+)
+
+// SecurityException CRDs are served at kubescape.io/v1beta1 (see the Helm chart
+// and kubescape's CRDExceptionsGetter). Keep this in sync when the CRD graduates.
+const (
+	securityExceptionGroup   = "kubescape.io"
+	securityExceptionVersion = "v1beta1"
+)
+
+var (
+	securityExceptionGVR = schema.GroupVersionResource{
+		Group:    securityExceptionGroup,
+		Version:  securityExceptionVersion,
+		Resource: "securityexceptions",
+	}
+	clusterSecurityExceptionGVR = schema.GroupVersionResource{
+		Group:    securityExceptionGroup,
+		Version:  securityExceptionVersion,
+		Resource: "clustersecurityexceptions",
+	}
+)
+
+// Tunables declared as vars (not consts) so tests can shorten them.
+var (
+	// securityExceptionCooldown debounces rapid changes to the same exception.
+	securityExceptionCooldown = 5 * time.Second
+	// securityExceptionEvictionInterval is how often the cooldown cache is swept.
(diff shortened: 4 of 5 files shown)


matthyx (commented):
Thanks for this — the design writeup, the coalescing/throttle model, the `-race` tests and the E2E against a live kind cluster are all excellent, and closing the GitOps loop on the runtime side is genuinely useful. I don't see a hard merge blocker; the change is safe by construction (idempotent full-cluster rescan, no status writes, no new RBAC/transport). A few things I'd like addressed or at least explicitly acknowledged before merge, one of them more important than the rest.

### 1. Watch resilience — list-once + bare re-watch can silently miss changes (most important)

`watchKind` calls `listExisting` exactly once, then loops calling `dynamicClient.Resource(gvr)...Watch(ctx, metav1.ListOptions{})` with no `ResourceVersion` and no relist on reconnect (`watcher/securityexceptionwatcher.go`, `watchKind`/`consumeWatch`).

A watch channel closing mid-run is normal (apiserver rotates watches every few minutes). Because you neither carry a `resourceVersion` across the reconnect nor relist, any create/update/delete that lands in the gap between the old watch closing and the new `Watch()` being established is lost — nothing else recovers it (the expiry sweep only reacts to `expiresAt`, not to general changes). The failure mode is subtle: an exception applied at just the wrong moment never triggers a rescan, and the GitOps loop this PR is meant to close quietly doesn't fire.

Rather than hand-rolling list+watch, please use client-go machinery that handles this: a dynamic informer (`dynamicinformer.NewDynamicSharedInformerFactory`) or `cache.NewListWatchFromClient` + `watchtools.NewRetryWatcher`, which track `resourceVersion` and relist on "resource version too old". That also removes the manual `watch.Bookmark`/`watch.Error`/retry handling. This is the one item I'd consider close-to-blocking, because the whole point of the feature is not missing changes.

### 2. Rescan is dropped on dispatch error

In `rescanLoop`, if `dispatchRescan` returns an error (e.g. the shared `eventWorkerPool` is overloaded and `Invoke` returns `ErrPoolOverload`) the error is logged and the pending rescan is discarded — the change isn't reflected until the next unrelated event or the scheduled scan. Consider re-arming the signal on failure (`wh.requestRescan()` in the error branch) so a transient pool-overload retries after the throttle rather than being silently lost.

### 3. Please confirm the shipped RBAC before merge

The description says no new RBAC is needed because the operator already has `get`/`list`/`watch` on these CRDs from helm-charts#875. Worth double-checking the deployed `ClusterRole` actually covers both `securityexceptions` and `clustersecurityexceptions` at `kubescape.io` — if it doesn't, `watchKind` logs an error and retries every 5s forever with zero rescans and no other signal, i.e. a silent no-op. A cheap way to make that non-silent would be to surface a clearer/rate-limited error after N consecutive watch failures.

### Minor / nits (non-blocking)

- **Shutdown leak:** `eventQueue` (`CooldownQueue`) is never `Stop()`ed, and its eviction callback sends on an unbuffered channel while `handleEvents` has already returned on `ctx.Done()`, so the cache goroutine can block permanently. Harmless for a process-lifetime component, but a `defer wh.eventQueue.Stop()` in `SecurityExceptionWatch` would be tidy.
- **List consistency:** `listExisting` pages via `pager.New`, but `sweepExpired` uses a plain unpaginated `List`. Fine at expected cardinalities; worth matching for consistency if the exception count can grow.
- The `v1beta1` constant and the full-cluster-vs-targeted-rescan tradeoff are already called out in the description and I agree with both choices as-is — good follow-up material, not merge concerns.

Nice work overall; addressing #1 (and ideally #2) would make this solidly mergeable.

yugal07:
Thanks, Matthias, all three addressed.

 Replaced the hand-rolled list-once + bare re-watch with a dynamic informer per kind, so resourceVersion tracking and relist-on-reconnect are handled by client-go and no change is missed across watch rotations (this also removed the manual Bookmark/Error handling).
 RescanLoop now re-arms the signal on a dispatch error so a transient pool overload retries after the throttle instead of being dropped.
 Good catch on the RBAC - the operator ClusterRole only grants get/list/watch on these CRDs when capabilities.riskAcceptance=enable, but the watcher was gating on Kubescape.Enabled, so a default install (riskAcceptance disabled) started the watcher with no RBAC; I plumbed riskAcceptance into the operator config and now gate the watcher on RiskAcceptanceEnabled() && Kubescape.Enabled so it exactly matches the chart. Verified against a live kind cluster: create/update/delete each dispatch a rescan, an injected dispatch error still results in a retried rescan, and helm template confirms the SE rule is present only with riskAcceptance enabled, matching the new gate. As a non-silent safety net, a persistently failing watch now escalates to an RBAC hint after N consecutive failures.

matthyx (commented):
Re-reviewed after `fdca18d` — all three points from my previous review are properly addressed, and cleanly. Thanks for the quick turnaround.

1. **Watch resilience ✅** — replaced the hand-rolled list-once + bare re-watch with a `dynamicinformer.NewFilteredDynamicSharedInformerFactory` per kind. The shared reflector tracks `resourceVersion` and relists on watch errors, so changes across watch rotations are no longer leaked, and the initial cache sync replays pre-existing exceptions as Adds. Nice touch handling `DeletedFinalStateUnknown` tombstones in `enqueueInformerEvent`.
2. **Dropped rescan on dispatch error ✅** — `rescanLoop` now re-arms via `requestRescan()` on dispatch failure, with the throttle sleep preventing a tight retry loop.
3. **RBAC ✅** — gated on `RiskAcceptanceEnabled()` (matching the Helm chart's conditional RBAC grant) *and* `Kubescape.Enabled`, plus a `WatchErrorHandler` that escalates to an explicit "verify the ClusterRole grants get/list/watch…" log after 3 consecutive failures — so a misconfigured Role is no longer a silent no-op. The config accessor has its own unit test too.

Both minor nits are handled as well: `defer wh.eventQueue.Stop()` in `SecurityExceptionWatch`, and `sweepExpired` now paginates via `pager.New`.

One note on the two open CodeRabbit comments: the "`sweepExpired` holds `wh.mu` across Kube API list calls" one is already resolved in this revision — the paginated List builds `entries` with no lock held, and `wh.mu` is only taken around the in-memory map update afterward. The `watch.Error` logging one is now moot since the informer's `WatchErrorHandler` covers that path.

CI is green (build, vet, cross-platform, basic tests, CodeQL all pass) and the new tests cover the live-create informer path, the re-arm, and the RBAC accessor. LGTM — no remaining blockers from my side.

matthyx (approved):

PR #396, opened 2026-07-22
fix: look up ContainerProfile by with-container slug, not workload slug

### LINKED ISSUE #395: Scheduled scan never triggers container-profile scans: ContainerProfile looked up by ApplicationProfile slug
## Summary

The scheduled vulnerability scan (`kubevuln-scheduler` CronJob) never triggers container-profile (relevancy) scans. Every container silently falls back to a plain image scan.

The cause is that `ContainerProfile` objects are looked up by the **no-container** slug (`InstanceID.GetSlug(true)`), which is the *ApplicationProfile* naming convention. `ContainerProfile` objects are named with the **with-container** slug (`InstanceID.GetSlug(false)`), so the lookup always returns `NotFound`.

## Affected code

`mainhandler/handlerequests.go` — `HandleImageScanningScopedRequest`:

```go
noContainerSlug, _ := instanceID.GetSlug(true)          // :429  <-- AP naming
...
if profile := utils.GetContainerProfileForRelevancyScan(ctx, mainHandler.ksStorageClient, noContainerSlug, ns); profile != nil {   // :442
```

`utils.GetContainerProfileForRelevancyScan` (`utils/containerprofile.go:48`) does
`storageClient.SpdxV1beta1().ContainerProfiles(namespace).Get(ctx, slug, ...)`.

The same defect exists in `watcher/podwatcher.go` at lines 121 and 158 (`handlePodWatcher`), which uses the identical `noContainerSlug` lookup.

## Evidence

Naming convention on a running cluster:

```
ContainerProfile : replicaset-<workload>-<hash>-<containerName>-<h1>-<h2>   <- GetSlug(false)
ApplicationProfile: replicaset-<workload>-<hash>                            <- GetSlug(true)
```

```
$ kubectl -n <ns> get containerprofiles.spdx.softwarecomposition.kubescape.io replicaset-<workload>-<hash>
Error from server (NotFound): containerprofiles.spdx.softwarecomposition.kubescape.io "replicaset-<workload>-<hash>" not found

$ kubectl -n <ns> get containerprofiles.spdx.softwarecomposition.kubescape.io replicaset-<workload>-<hash>-<containerName>-<h1>-<h2> -o jsonpath='{.metadata.annotations}'
{"kubescape.io/completion":"complete","kubescape.io/instance-id":"...","kubescape.io/status":"completed","kubescape.io/wlid":"wlid://...", ...}
```

The second object has `status: completed` plus valid `instance-id` and `wlid` annotations, i.e. it would pass `utils.SkipContainerProfile` and *should* have produced a container-profile scan.

Operator logs on an affected cluster show **zero** occurrences of `"found container profile"` or `"container profile not found"` — `GetContainerProfileForRelevancyScan` has never returned a hit. All observed `"scanning container profile"` lines come from a different code path (see below).

## Why container-profile scans still appear to work at startup

There are two independent entry points that traverse in opposite directions:

**Profile-driven (works)** — `watcher/containerprofilewatcher.go`. `ContainerProfileWatch` pages through all `ContainerProfile` objects and enqueues each as a synthetic `watch.Added`; `HandleContainerProfileEvents` reads `obj.Name` / `obj.Namespace` / `obj.Annotations[wlid]` directly off the object. No slug is derived and no `Get()` by name is performed, so nothing can go wrong. This produces the burst of `"scanning container profile"` log lines at operator startup, and reacts to subsequent CP add/modify events.

**Pod-driven (broken)** — the CronJob POSTs
`{"commands":[{"commandName":"scan","designators":[{"designatorType":"Attributes","attributes":{}}]}]}`
to `operator:4002/v1/triggerAction`. Since `apis.TypeScanImages == "scan"` and `Designators` is non-empty, `handleRequest` sets `isToItemizeScopeCommand = true` and dispatches to `HandleImageScanningScopedRequest`, which walks namespaces → pods → containers and *derives* a slug to look the profile up. That derivation is the step the watcher never has to perform, and it is where the wrong slug variant is used.

`watcher/podwatcher.go` is also pod-driven, hence the same bug.

## Impact

The `ContainerProfile` watcher only fires on CP add/modify plus the one-time startup list. The daily CronJob exists to re-scan everything against a refreshed CVE database — and each of those runs currently downgrades a relevancy scan to a plain image scan. Relevancy results are therefore progressively overwritten with non-relevancy ones until the operator restarts or node-agent touches the profile.

## Second, latent bug

`mainhandler/handlerequests.go:429-433`, introduced in 8f670a8 ("fix: skip scanning remaining containers for already processed profiles"):

```go
noContainerSlug, _ := instanceID.GetSlug(true)
if ok := slugs[noContainerSlug]; ok {
    // container profile already scanned for this workload, skip remaining containers
    continue
}
```

with `slugs[noContainerSlug] = true` set after a successful profile scan. This dedup was correct when profiles were one-per-**workload** (`ApplicationProfile`), but `ContainerProfile` is one-per-**container**. Once the slug bug above is fixed, this would silently drop every container after the first in multi-container workloads. On one test cluster that is 49 of 616 eligible ContainerProfiles across 40 multi-container workloads.

`watcher/podwatcher.go` has the same pattern via its `noContainerSlugs` map (lines 113-118 and 150-155).

## Regression origin

5ad0c5b ("trigger AP scan on ContainerProfiles", 2025-09-29) migrated `GetApplicationProfileForRelevancyScan` → `GetContainerProfileForRelevancyScan`, switching the queried resource from `ApplicationProfiles` to `ContainerProfiles`, but kept passing the AP-style (no-container) name.

## Proposed fix

Minimal correction: pass the with-container slug (already computed as `s` at `handlerequests.go:423`) and remove the `noContainerSlug` dedup — the existing `slugs[s]` check already provides per-container dedup. Mirror in `podwatcher.go`.

Worth considering instead: make the pod-driven path resolve profiles the way the watcher does, rather than re-deriving names. Concretely, `List` the `ContainerProfile` objects in scope once up front and build an index, then match pod containers against it — keyed on the `kubescape.io/instance-id` annotation (exact, and immune to the length-truncation rules in `names.InstanceIDToSlugWithMaxLength`) or on the workload labels the CP already carries (`kubescape.io/workload-kind`, `workload-name`, `workload-namespace`, `workload-container-name`, `instance-template-hash`). The pod walk is still needed to emit plain image scans for containers that have no profile and to honor the label/field selectors, but no name would ever be derived by the operator again. This also replaces one `GET` per container per scan with a single `LIST`.

## Regression test

Add a case to `mainhandler/handlerequests_test.go` covering `HandleImageScanningScopedRequest` with a fake storage client holding a `ContainerProfile` named with `GetSlug(false)`, asserting a `CommandScanContainerProfile` is produced (not `TypeScanImages`), and a multi-container workload case asserting one command per container.

matthyx:
## Overview
This PR fixes #395

`ContainerProfile` objects are named using the with-container slug (`InstanceID.GetSlug(false)`), but `HandleImageScanningScopedRequest` and `handlePodWatcher` looked them up using the no-container slug (`GetSlug(true)`, the `ApplicationProfile` naming convention). The lookup therefore always missed, and every scheduled scan (the `kubevuln-scheduler` CronJob) silently fell back to a plain image scan instead of a relevancy (container-profile) scan.

Three fixes, all in this PR:

1. **`mainhandler/handlerequests.go`** — `HandleImageScanningScopedRequest` now passes the with-container slug (already computed as `s`) into `GetContainerProfileForRelevancyScan`, and dedups on `slugs[s]` instead of a separate no-container-slug map.
2. **`watcher/podwatcher.go`** — `handlePodWatcher` had the identical bug pattern at two call sites; same fix, using the outer loop's `slug` (already with-container).
3. **`utils/containerprofile.go`** — a second, previously-unreported bug found while verifying this fix: `GetContainerProfileScanCommand` built its command with `CommandName: apis.TypeScanApplicationProfile`, which has no case in `runCommand`'s dispatch switch (it only recognizes `utils.CommandScanContainerProfile`) — so even a correctly-found profile's scan command would have been silently dropped. Fixed to use the correct constant.

Both call sites also previously deduped container-profile scans per *workload* instead of per *container* (via a no-container-slug map), which — once the slug itself was fixed naively — would have dropped every container after the first in multi-container workloads. That dedup logic has been removed; the existing per-container slug dedup (`slugs[s]` / the outer per-container loop key) already provides correct dedup.

## How to Test
New regression tests were added and verified by mutation (fail against the pre-fix code, pass against the fix — confirmed by temporarily reverting each production change and rerunning):
- `watcher/podwatcher_test.go`: `Test_handlePodWatcher_ContainerProfile` — single-container hit case, and a multi-container case asserting one scan command per container (regression test for the removed per-workload dedup).
- `mainhandler/handlerequests_test.go`: `TestHandleImageScanningScopedRequest_LooksUpWithContainerSlug` and `TestHandleImageScanningScopedRequest_MultiContainerDoesNotDedupeAcrossContainers`. These enable Kubevuln against an `httptest.Server` stub so a correctly-dispatched scan completes end-to-end through `actionHandler.scanContainerProfile`, and assert on the stub's received request count — this is what actually distinguishes a correct fix from one that finds the profile but silently drops the resulting command (see fix #3) or dedups across containers (see the removed per-workload dedup).

`go build ./...`, `go vet ./...`, and `go test -count=1 ./mainhandler/... ./watcher/... ./utils/...` (and the rest of the repo's packages) pass. Note: `watcher/registryhandler_test.go`'s `TestRegistryCommandWatch` is a testcontainers-based integration test that is independently flaky and not exercised by this PR's branch (it's untracked locally); it is unrelated to this change and fails the same way on `main`.

## Additional Information
Out of scope for this PR: the issue also floats a larger redesign (replace per-container `Get()` calls with a single `List` matched on the `kubescape.io/instance-id` annotation, closing a remaining slug-truncation edge case) as a "worth considering instead" follow-up, not a requirement for this fix.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

AI-skills: oh-my-claudecode:team | cmds: /oh-my-claudecode:ralplan

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image scanning to correctly track and process individual containers within the same pod.
  * Fixed container profile lookups so scans use the appropriate per-container profile.
  * Prevented valid scans from being skipped when multiple containers share a workload.
  * Ensured container-profile scan commands are correctly identified and processed.
* **Tests**
  * Added coverage for single-container and multi-container scanning scenarios.

### DIFF
--- a/mainhandler/handlerequests.go
+++ b/mainhandler/handlerequests.go
@@ -426,12 +426,6 @@ func (mainHandler *MainHandler) HandleImageScanningScopedRequest(ctx context.Con
 					continue
 				}
 
-				noContainerSlug, _ := instanceID.GetSlug(true)
-				if ok := slugs[noContainerSlug]; ok {
-					// container profile already scanned for this workload, skip remaining containers
-					continue
-				}
-
 				// get container data
 				containerData, err := utils.PodToContainerData(mainHandler.k8sAPI, pod, instanceID, mainHandler.config.ClusterName())
 				if err != nil {
@@ -439,7 +433,7 @@ func (mainHandler *MainHandler) HandleImageScanningScopedRequest(ctx context.Con
 					continue
 				}
 
-				if profile := utils.GetContainerProfileForRelevancyScan(ctx, mainHandler.ksStorageClient, noContainerSlug, ns); profile != nil {
+				if profile := utils.GetContainerProfileForRelevancyScan(ctx, mainHandler.ksStorageClient, s, ns); profile != nil {
 					cmd := utils.GetContainerProfileScanCommand(profile, pod)
 
 					// send specific command to the channel
@@ -451,7 +445,7 @@ func (mainHandler *MainHandler) HandleImageScanningScopedRequest(ctx context.Con
 						continue
 					}
 					logger.L().Info("action completed successfully", helpers.String("name", profile.Name), helpers.String("namespace", profile.Namespace))
-					slugs[noContainerSlug] = true
+					slugs[s] = true
 				} else {
 					// set scanning command
 					cmd := &apis.Command{
--- a/mainhandler/handlerequests_test.go
+++ b/mainhandler/handlerequests_test.go
@@ -3,12 +3,18 @@ package mainhandler
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
+	utilsmetadata "github.com/armosec/utils-k8s-go/armometadata"
+
 	"github.com/armosec/armoapi-go/apis"
 	"github.com/armosec/armoapi-go/identifiers"
+	beUtils "github.com/kubescape/backend/pkg/utils"
 	instanceidhandlerv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1"
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 	"github.com/kubescape/operator/config"
@@ -146,15 +152,61 @@ func getContainerProfileActionNames(t *testing.T, storageClient interface{ Actio
 	return names
 }
 
+// scannerStub stands in for kubevuln. HandleSingleRequest only reaches it if the dispatch
+// switch in runCommand actually routes the command to actionHandler.scanContainerProfile
+// (i.e. GetContainerProfileScanCommand set the right CommandName) and that handler completes
+// successfully (which is also the only way slugs[s] = true gets set in
+// HandleImageScanningScopedRequest). Counting requests therefore lets tests detect both the
+// CommandName regression and the per-workload dedup regression, neither of which a bare
+// "get containerprofiles" assertion on the storage client can distinguish from a correct fix.
+type scannerStub struct {
+	server *httptest.Server
+	mu     sync.Mutex
+	count  int
+}
+
+func newScannerStub(t *testing.T) *scannerStub {
+	t.Helper()
+	s := &scannerStub{}
+	s.server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		s.mu.Lock()
+		s.count++
+		s.mu.Unlock()
+		w.WriteHeader(http.StatusOK)
+	}))
+	t.Cleanup(s.server.Close)
+	return s
+}
+
+func (s *scannerStub) requestCount() int {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.count
+}
+
+func (s *scannerStub) hostPort() string {
+	return strings.TrimPrefix(s.server.URL, "http://")
+}
+
 // newMainHandlerForTest builds a MainHandler wired to fake Kubernetes and storage clientsets,
-// with Kubevuln disabled (so any container-profile-found path that reaches
-// actionHandler.scanContainerProfile returns a harmless error immediately, with no further
-// network calls).
-func newMainHandlerForTest(k8sClient *k8sfake.Clientset, storageClient *kssfake.Clientset) *MainHandler {
+// with Kubevuln enabled and pointed at scanner. This lets a found ContainerProfile's scan
+// command actually complete end-to-end through actionHandler.scanContainerProfile, instead of
+// short-circuiting on a "kubevuln is not enabled" error before slugs[s] = true is ever reached.
+func newMainHandlerForTest(k8sClient *k8sfake.Clientset, storageClient *kssfake.Clientset, scanner *scannerStub) *MainHandler {
+	capabilities := config.CapabilitiesConfig{
+		Components: config.Components{
+			Kubevuln: config.Component{Enabled: true},
+		},
+	}
+	clusterConfig := utilsmetadata.ClusterConfig{KubevulnURL: scanner.hostPort()}
+	// VulnScanHttpClient is normally set once at startup (see main.go); scanContainerProfile
+	// uses it directly (not via config), so tests that expect a real POST to reach the
+	// scanner stub must set it explicitly too.
+	VulnScanHttpClient = utils.InitHttpClient(clusterConfig.KubevulnURL)
 	return &MainHandler{
 		k8sAPI:          utils.NewK8sInterfaceFake(k8sClient),
 		ksStorageClient: storageClient,
-		config:          newTestConfig(config.Config{}),
+		config:          config.NewOperatorConfig(capabilities, clusterConfig, &beUtils.Credentials{}, config.Config{}),
 	}
 }
 
@@ -179,16 +231,20 @@ func TestHandleImageScanningScopedRequest_LooksUpWithContainerSlug(t *testing.T)
 	slugs := withContainerSlugsForHandler(t, pod)
 	require.Len(t, slugs, 1)
 
-	k8sClient := k8sfake.NewSimpleClientset(pod)
+	k8sClient := k8sfake.NewClientset(pod)
 	storageClient := kssfake.NewSimpleClientset(containerProfileForHandler(ns, slugs[0]))
+	scanner := newScannerStub(t)
 
-	mainHandler := newMainHandlerForTest(k8sClient, storageClient)
+	mainHandler := newMainHandlerForTest(k8sClient, storageClient, scanner)
 	sessionObj := sessionObjForNamespaceForHandler(ns)
 
 	mainHandler.HandleImageScanningScopedRequest(context.Background(), sessionObj)
 
 	gotNames := getContainerProfileActionNames(t, storageClient)
 	assert.Equal(t, slugs, gotNames, "expected a single 'get containerprofiles' call using the with-container slug")
+	// Only reachable if GetContainerProfileScanCommand's CommandName correctly routes through
+	// runCommand's dispatch switch into actionHandler.scanContainerProfile.
+	assert.Equal(t, 1, scanner.requestCount(), "expected the container-profile scan to actually dispatch to the scanner")
 }
 
 // TestHandleImageScanningScopedRequest_MultiContainerDoesNotDedupeAcrossContainers is a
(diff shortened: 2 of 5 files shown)


matthyx (commented):
## The production fix is correct

All three changes verified:

1. **Slug fix** — `mapSlugToInstanceID` (`podwatcher.go:230`) and the `slugsToImages` map both key on `GetSlug(false)`, so passing `slug` / `s` is right, and it matches the actual `ContainerProfile` naming.
2. **`CommandName` fix** — real bug, and the more severe of the two. Confirmed by printing the constants: `apis.TypeScanApplicationProfile == "scanApplicationProfile"` while `utils.CommandScanContainerProfile == "scanContainerProfile"`, and `runCommand`'s switch only cases the latter — so the command fell through to `default:` and was dropped. Note that `scanContainerProfile` still hardcodes `apis.TypeScanApplicationProfile` at `vulnscan.go:303` for scanner-URL selection; that is correct (it is the kubevuln endpoint path) and rightly left untouched.
3. **Dedup removal** — safe. The `SlugToImageID.Set` / `WlidAndImageID.Add` calls that lived inside the deleted skip-branches are duplicated in the fall-through tail, so no cache state is lost.

I checked the fix direction by mutation rather than by reading: reverting the slug fix makes both new `mainhandler` tests fail, and reverting the `CommandName` fix makes the new `podwatcher` test fail.

## Two test-quality problems, both proven by mutation

### 1. `TestHandleImageScanningScopedRequest_MultiContainerDoesNotDedupeAcrossContainers` does not test what it claims

I restored the per-workload dedup exactly as it was before this PR (re-adding the `noContainerSlug` lookup/`continue` and `slugs[noContainerSlug] = true`) while keeping the corrected slug in the profile lookup — and the test still passed.

Cause: `newMainHandlerForTest` disables Kubevuln, so `HandleSingleRequest` returns `"kubevuln is not enabled"` and the `continue` at `handlerequests.go:444` fires *before* `slugs[...] = true` at `:448`. The dedup marker is therefore never set, and the dedup can never trigger inside this test. The test log confirms all three containers take the error path:

```
[info]  triggering container profile scan. name: pod-my-pod-nginx-d00d-dd12
[error] failed to complete action. error: kubevuln is not enabled; ...
[info]  triggering container profile scan. name: pod-my-pod-sidecar-7ab1-e32f
[error] failed to complete action. error: kubevuln is not enabled; ...
[info]  triggering container profile scan. name: pod-my-pod-init-66cb-5b76
[error] failed to complete action. error: kubevuln is not enabled; ...
```

As written it is effectively a duplicate of `TestHandleImageScanningScopedRequest_LooksUpWithContainerSlug` — it covers the slug, not the dedup.

### 2. Neither `mainhandler` test covers fix #3

Reverting only the `utils/containerprofile.go` change (`CommandScanContainerProfile` → `apis.TypeScanApplicationProfile`), leaving both slug fixes in place, leaves `go test ./mainhandler/...` green. Only the `podwatcher` test catches it — which it does, correctly.

So the scoped-request path, which is the one #395 is actually about, has no assertion that it emits a `CommandScanContainerProfile`.

### Suggested fix for both

Both findings stem from the same choice: the `mainhandler` tests assert on recorded fake-client `get containerprofiles` actions rather than on the emitted command. Enabling Kubevuln in the test config and pointing `KubevulnURL` at an `httptest.NewServer` addresses both at once — `HandleSingleRequest` then succeeds, so the dedup marker actually gets set (making finding 1 detectable) and dispatch has to survive `runCommand`'s switch (making finding 2 detectable). The new `podwatcher` tests already assert on emitted commands and are the better model here.

## Minor

The description states that `go test -count=1 ./...` passes across the full repo. `TestRegistryCommandWatch` fails under `./watcher/...`. I checked out `main` and it fails there too, so it is pre-existing and not caused by this PR — but the claim should be corrected so the failure is not mistaken for a regression from these changes.

matthyx:
Thanks both for the review — pushed 70f45ca addressing this.

**CodeRabbit nitpick**: replaced the deprecated `k8sfake.NewSimpleClientset` with `k8sfake.NewClientset` in the two new call sites in `mainhandler/handlerequests_test.go`.

**@matthyx — both test-quality findings were real, fixed by the same change**: `newMainHandlerForTest` now enables Kubevuln and points `KubevulnURL` at an `httptest.Server` stub, so a correctly-dispatched scan completes end-to-end through `actionHandler.scanContainerProfile` instead of short-circuiting on `"kubevuln is not enabled"` before `slugs[s] = true` is ever reached. Both tests now additionally assert on the stub's received request count.

Verified by mutation, the same way you found it:
- Reverting the `CommandName` fix in `utils/containerprofile.go` → both tests now fail (0 requests reach the stub instead of 1/3).
- Reintroducing the old per-workload dedup (keyed on the no-container slug) in `handlerequests.go` → the multi-container test now fails (1 request instead of 3).

Also: you're right that the PR description's "`go test -count=1 ./...` passes across the full repo" claim was made without `watcher/registryhandler_test.go` in the picture — that file isn't part of this PR (it's untracked locally), and `TestRegistryCommandWatch` is a testcontainers-based integration test that's independently flaky/pre-existing on `main`, unrelated to this change. Updating the PR description to not overstate that.

matthyx (commented):
Re-reviewed at 70f45ca. Production code is unchanged from the previous round; this commit is test-only. Both earlier findings are addressed, and I verified the mutation claims independently rather than taking the commit message's word for them:

- **Revert the `CommandName` fix** → `TestHandleImageScanningScopedRequest_LooksUpWithContainerSlug` fails (`expected: 1, actual: 0`) and `..._MultiContainerDoesNotDedupeAcrossContainers` fails (`expected: 3, actual: 0`). Previously both stayed green.
- **Reintroduce the per-workload dedup** → the multi-container test fails (`expected: 3, actual: 1`). Previously it stayed green.

Also confirmed: `go vet` clean on the touched packages, `./mainhandler/...` and `./utils/...` pass, and the two new tests pass under `-race` (the stub's counter is mutex-guarded and the handler goroutines are the only writers). `TestRegistryCommandWatch` still fails under `./watcher/...`, unchanged and pre-existing on `main`.

Two minor points, neither blocking:

### 1. `VulnScanHttpClient` is mutated globally without restore

`newMainHandlerForTest` assigns the package-level `VulnScanHttpClient` (`handlerequests.go:62`) and never restores it, while the `httptest.Server` it points at is closed by `t.Cleanup`. Nothing else in the package currently reaches that global, so this is harmless today — but it is a latent trap. A future `mainhandler` test that reaches `sendWorkloadWithCredentials` after these two run would POST to a closed port and hit the connection-refused retry loop at `vulnscan.go:448`: 5 retries × a 5s sleep, i.e. a ~25s hang that would present as a mysteriously slow test rather than a clear failure.

Cheap guard:

```go
prev := VulnScanHttpClient
t.Cleanup(func() { VulnScanHttpClient = prev })
VulnScanHttpClient = utils.InitHttpClient(clusterConfig.KubevulnURL)
```

(which requires threading `t` into the helper).

### 2. The scanner stub counts requests but does not discriminate by path

The stub accepts every path, so the request count proves *a* scan dispatched, not that it was a *container-profile* scan. Verified by mutation: reverting only the slug fix leaves both count assertions satisfied (1 and 3), because the fallback image-scan path POSTs to the same stub — only the `getContainerProfileActionNames` assertion catches that regression.

The two assertions together are adequate, so coverage is not actually missing. But asserting the path in the stub handler (`/v1/scanApplicationProfile` for the container-profile route, per `getAPScanURL`, vs. the image-scan route from `getVulnScanURL`) would make the count assertion mean what its comment claims — "the container-profile scan actually dispatched" — instead of "some scan dispatched".

LGTM otherwise.

PR #402, opened 2026-08-03
fix(rulebinding): match namespaces by exact name, not String() substring

### LINKED ISSUE #401: admission/rulebinding: NamespaceSelector match uses strings.Contains on NamespaceList.String()
# Description
Admission rule-binding `NamespaceSelector` matching uses `strings.Contains(namespaces.String(), ns)` in `admission/rulebinding/cache/cache.go`. That substring-checks a debug dump, so e.g. namespace `dev` wrongly matches a binding that selected `devel`.

# Environment
OS: `mac os`
Version: operator (current main)

# Steps To Reproduce
1. Create a `RuntimeAlertRuleBinding` whose `namespaceSelector` matches only `devel`.
2. Create/update an object in namespace `dev` that matches `podSelector`.
3. Binding still applies (false positive).

# Expected behavior
Only exact namespace names from the selected list match.

# Actual Behavior
Substring match on `NamespaceList.String()` → wrong namespaces get admission rules/alerts.

manumathon:
## Overview
fixes #401
This PR fixes false-positive rulebinding matches when a namespace name is a substring of another (e.g. `dev` matching `devel`).

Namespace membership now checks exact names instead of `strings.Contains(namespaces.String(), …)`. Invalid label selectors are skipped.

## How to Test
1. Label `devel` with `repro=selected`; leave `dev` unlabeled.
2. Bind rule `R2002` with that namespace selector.
3. Create pods in both namespaces — only `devel` should match.

- [ ] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rule-binding label selector handling by reporting conversion errors and skipping invalid bindings.
  * Fixed namespace matching to require exact names, preventing unintended matches from partial or substring values.
  * Added safer handling for empty or unavailable namespace lists.

### DIFF
--- a/admission/rulebinding/cache/cache.go
+++ b/admission/rulebinding/cache/cache.go
@@ -2,7 +2,6 @@ package cache
 
 import (
 	"context"
-	"strings"
 
 	"github.com/goradd/maps"
 	"github.com/kubescape/go-logger"
@@ -96,14 +95,22 @@ func (c *RBCache) ListRulesForObject(ctx context.Context, object *unstructured.U
 		}
 
 		// check pod selectors
-		podSelector, _ := metav1.LabelSelectorAsSelector(&rb.Spec.PodSelector)
+		podSelector, err := metav1.LabelSelectorAsSelector(&rb.Spec.PodSelector)
+		if err != nil {
+			logger.L().Error("failed to parse pod selector", helpers.String("ruleBiding", uniqueName(&rb)), helpers.Error(err))
+			continue
+		}
 		if !podSelector.Matches(labels.Set(object.GetLabels())) {
 			// pod selectors doesnt match
 			continue
 		}
 
 		// check namespace selectors
-		nsSelector, _ := metav1.LabelSelectorAsSelector(&rb.Spec.NamespaceSelector)
+		nsSelector, err := metav1.LabelSelectorAsSelector(&rb.Spec.NamespaceSelector)
+		if err != nil {
+			logger.L().Error("failed to parse namespace selector", helpers.String("ruleBiding", uniqueName(&rb)), helpers.Error(err))
+			continue
+		}
 		nsSelectorStr := nsSelector.String()
 		if len(nsSelectorStr) != 0 {
 			// get related namespaces
@@ -112,7 +119,7 @@ func (c *RBCache) ListRulesForObject(ctx context.Context, object *unstructured.U
 				logger.L().Error("failed to list namespaces", helpers.String("ruleBiding", uniqueName(&rb)), helpers.String("nsSelector", nsSelectorStr), helpers.Error(err))
 				continue
 			}
-			if !strings.Contains(namespaces.String(), object.GetNamespace()) {
+			if !namespaceListHasName(namespaces, object.GetNamespace()) {
 				// namespace selectors dont match
 				continue
 			}
--- a/admission/rulebinding/cache/helpers.go
+++ b/admission/rulebinding/cache/helpers.go
@@ -4,10 +4,26 @@ import (
 	typesv1 "github.com/kubescape/node-agent/pkg/rulebindingmanager/types/v1"
 	"github.com/kubescape/node-agent/pkg/utils"
 	"github.com/kubescape/node-agent/pkg/watcher"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 )
+
+// namespaceListHasName reports whether name is an exact member of list.Items.
+// Do not use strings.Contains(list.String(), name): NamespaceList.String() is a
+// debug dump, so substrings (e.g. "dev" in "devel") falsely match.
+func namespaceListHasName(list *corev1.NamespaceList, name string) bool {
+	if list == nil {
+		return false
+	}
+	for i := range list.Items {
+		if list.Items[i].Name == name {
+			return true
+		}
+	}
+	return false
+}
 
 // ruleBindingKind is the Kind value the CRD reports for RuntimeAlertRuleBinding
 // objects. The dynamic watcher dispatches events from all watched GVRs to every
--- a/admission/rulebinding/cache/helpers_test.go
+++ b/admission/rulebinding/cache/helpers_test.go
@@ -10,6 +10,25 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
+func TestNamespaceListHasName(t *testing.T) {
+	list := &corev1.NamespaceList{
+		Items: []corev1.Namespace{
+			{ObjectMeta: metav1.ObjectMeta{Name: "devel"}},
+			{ObjectMeta: metav1.ObjectMeta{Name: "production"}},
+		},
+	}
+
+	assert.True(t, namespaceListHasName(list, "devel"))
+	assert.True(t, namespaceListHasName(list, "production"))
+	// Substring of "devel" / "production" must not match (regression for
+	// strings.Contains(namespaces.String(), ns)).
+	assert.False(t, namespaceListHasName(list, "dev"))
+	assert.False(t, namespaceListHasName(list, "prod"))
+	assert.False(t, namespaceListHasName(list, "missing"))
+	assert.False(t, namespaceListHasName(nil, "devel"))
+	assert.False(t, namespaceListHasName(&corev1.NamespaceList{}, "devel"))
+}
+
 func TestResourcesToWatch(t *testing.T) {
 	tests := []struct {
 		name string


admission/rulebinding/cache/cache.go
```diff
@@ -96,14 +95,22 @@ func (c *RBCache) ListRulesForObject(ctx context.Context, object *unstructured.U
 		}
 
 		// check pod selectors
-		podSelector, _ := metav1.LabelSelectorAsSelector(&rb.Spec.PodSelector)
+		podSelector, err := metav1.LabelSelectorAsSelector(&rb.Spec.PodSelector)
+		if err != nil {
+			logger.L().Error("failed to parse pod selector", helpers.String("ruleBiding", uniqueName(&rb)), helpers.Error(err))
```
matthyx, line 100: This is the most valuable change in the PR and the description doesn't mention it: on `main` these discarded errors are a **crash vector**, not a silent mismatch.

`metav1.LabelSelectorAsSelector` returns `(nil, err)` on a bad selector — it does *not* fall back to `labels.Nothing()`. So `podSelector.Matches(...)` on the old line dereferences a nil interface, and so does `nsSelector.String()` below. Verified on `main` with a `RuntimeAlertRuleBinding` whose `podSelector` uses an unknown operator:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]
github.com/kubescape/operator/admission/rulebinding/cache.(*RBCache).ListRulesForObject(...)
	admission/rulebinding/cache/cache.go:100
```

There is no `recover()` anywhere under `admission/` or in `main.go` (only `mainhandler` and `restapihandler` have one), and `ListRulesForObject` is called from the admission worker in `admission/webhook/validator.go:282` — so one malformed CRD, which the API server will happily accept since the CRD schema doesn't validate operator values, takes down the operator process. With this branch the same input logs `failed to parse pod selector ... "Bogus" is not a valid label selector operator` and skips the binding (fail-closed, which is the right call for an admission matcher).

Two asks:
- Mention this in the PR body / release notes — "fixes a panic that crashes the operator on an invalid rule binding selector" is a much bigger headline than "reports conversion errors".
- Add a test for it. It's ~20 lines and needs no fixtures:

```go
func TestListRulesForObjectInvalidSelectorDoesNotPanic(t *testing.T) {
	c := NewCacheMock()
	c.addRuleBinding(&typesv1.RuntimeAlertRuleBinding{
		ObjectMeta: metav1.ObjectMeta{Name: "badRB", Namespace: "kubescape"},
		Spec: typesv1.RuntimeAlertRuleBindingSpec{
			PodSelector: metav1.LabelSelector{
				MatchExpressions: []metav1.LabelSelectorRequirement{
					{Key: "app", Operator: "Bogus", Values: []string{"x"}},
				},
			},
			Rules: []typesv1.RuntimeAlertRuleBindingRule{{RuleID: "R2000"}},
		},
	})

	obj := &unstructured.Unstructured{Object: map[string]interface{}{
		"metadata": map[string]interface{}{
			"name": "p", "namespace": "ns",
			"labels": map[string]interface{}{"app": "x"},
		},
	}}

	assert.NotPanics(t, func() {
		assert.Empty(t, c.ListRulesForObject(context.Background(), obj))
	})
}
```

(Panics on `main`, passes on this branch.)

admission/rulebinding/cache/cache.go
```diff
@@ -96,14 +95,22 @@ func (c *RBCache) ListRulesForObject(ctx context.Context, object *unstructured.U
...
 		if !podSelector.Matches(labels.Set(object.GetLabels())) {
 			// pod selectors doesnt match
 			continue
 		}
 
 		// check namespace selectors
-		nsSelector, _ := metav1.LabelSelectorAsSelector(&rb.Spec.NamespaceSelector)
+		nsSelector, err := metav1.LabelSelectorAsSelector(&rb.Spec.NamespaceSelector)
+		if err != nil {
+			logger.L().Error("failed to parse namespace selector", helpers.String("ruleBiding", uniqueName(&rb)), helpers.Error(err))
```
matthyx, line 111: Two nits on both new log lines (this one and the pod-selector one above):

- `rbName` is already computed from `uniqueName(&rb)` at the top of the loop (`cache.go:81`) — reuse it instead of recomputing.
- `"ruleBiding"` is a typo. It's pre-existing on the `failed to list namespaces` line below, so it's consistent — but if it's being copied into two new lines, worth fixing all three to `"ruleBinding"` in the same commit. Anyone grepping logs or building a dashboard on this field will thank you.

```go
logger.L().Error("failed to parse namespace selector", helpers.String("ruleBinding", rbName), helpers.Error(err))
```

admission/rulebinding/cache/cache.go
```diff
@@ -112,7 +119,7 @@ func (c *RBCache) ListRulesForObject(ctx context.Context, object *unstructured.U
 				logger.L().Error("failed to list namespaces", helpers.String("ruleBiding", uniqueName(&rb)), helpers.String("nsSelector", nsSelectorStr), helpers.Error(err))
 				continue
 			}
-			if !strings.Contains(namespaces.String(), object.GetNamespace()) {
+			if !namespaceListHasName(namespaces, object.GetNamespace()) {
```
matthyx, line 122: **Must-fix: this call site — the one #401 is actually about — has no test.**

`TestNamespaceListHasName` tests the helper in isolation, so it would stay green if this line were reverted to `strings.Contains(namespaces.String(), object.GetNamespace())`. Nothing in the suite exercises `ListRulesForObject` with a non-empty `namespaceSelector` at all, which is exactly why the bug survived this long.

The existing `k8sinterface.NewKubernetesApiMock()` backs `KubernetesClient` with a real `kubernetesfake.NewSimpleClientset()`, and the generated fake's `Namespaces().List` does apply the label selector — so this reproduces the reported scenario end to end with no new test infrastructure:

```go
func TestListRulesForObjectNamespaceSelectorExactMatch(t *testing.T) {
	k8sAPI := k8sinterface.NewKubernetesApiMock()
	for _, ns := range []*corev1.Namespace{
		{ObjectMeta: metav1.ObjectMeta{Name: "devel", Labels: map[string]string{"repro": "selected"}}},
		{ObjectMeta: metav1.ObjectMeta{Name: "dev"}},
	} {
		_, err := k8sAPI.KubernetesClient.CoreV1().Namespaces().Create(context.Background(), ns, metav1.CreateOptions{})
		assert.NoError(t, err)
	}

	c := NewCacheMock()
	c.k8sClient = k8sAPI
	c.addRuleBinding(&typesv1.RuntimeAlertRuleBinding{
		ObjectMeta: metav1.ObjectMeta{Name: "testRB", Namespace: "kubescape"},
		Spec: typesv1.RuntimeAlertRuleBindingSpec{
			PodSelector:       metav1.LabelSelector{MatchLabels: map[string]string{"app": "testPod"}},
			NamespaceSelector: metav1.LabelSelector{MatchLabels: map[string]string{"repro": "selected"}},
			Rules:             []typesv1.RuntimeAlertRuleBindingRule{{RuleID: "R2000"}},
		},
	})

	pod := func(ns string) *unstructured.Unstructured {
		return &unstructured.Unstructured{Object: map[string]interface{}{
			"metadata": map[string]interface{}{
				"name": "testPod", "namespace": ns,
				"labels": map[string]interface{}{"app": "testPod"},
			},
		}}
	}

	assert.Len(t, c.ListRulesForObject(context.Background(), pod("devel")), 1, "selected namespace must match")
	assert.Len(t, c.ListRulesForObject(context.Background(), pod("dev")), 0, "substring namespace must NOT match")
}
```

I ran this both ways:

- on `main`: `Error: "[0xc000664810]" should have 0 item(s), but has 1 — substring namespace must NOT match`
- on this branch: `PASS`

Keep `TestNamespaceListHasName` too — it's good — but this is the one that pins the regression.

admission/rulebinding/cache/helpers.go
```diff
@@ -4,11 +4,27 @@ import (
...
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 )
 
+// namespaceListHasName reports whether name is an exact member of list.Items.
+// Do not use strings.Contains(list.String(), name): NamespaceList.String() is a
+// debug dump, so substrings (e.g. "dev" in "devel") falsely match.
+func namespaceListHasName(list *corev1.NamespaceList, name string) bool {
```
matthyx, line 16: Follow-up, not blocking this PR — the helper is the right fix for the bug as written.

But it's worth noting *why* this helper has to exist: the caller LISTs every namespace in the cluster matching the selector, then linearly scans the result for one name. That LIST is issued once per rule binding, on every admission request, in the webhook's latency path. With N bindings carrying a `namespaceSelector`, a single pod create costs N cluster-wide namespace LISTs against the API server.

The question being asked is "does *this one* namespace match the selector", so it can be answered without the fan-out:

```go
ns, err := c.k8sClient.GetKubernetesClient().CoreV1().Namespaces().Get(ctx, object.GetNamespace(), metav1.GetOptions{})
if err != nil {
	logger.L().Error("failed to get namespace", helpers.String("ruleBinding", rbName), helpers.Error(err))
	continue
}
if !nsSelector.Matches(labels.Set(ns.GetLabels())) {
	continue
}
```

One GET instead of a LIST (and ideally served from an informer/lister cache rather than the API server at all), no `nsSelector.String()` round-trip through `ListOptions`, and `namespaceListHasName` disappears — along with the whole class of bug it's guarding against. Same treatment would apply to node-agent's copy of this loop.

matthyx (commented):
Reviewed this against `main` in a local checkout of the branch: `go build ./...` and `go vet ./admission/...` are clean, and `go test ./...` is green (including `admission/rulebinding/cache`), so the unchecked "New and existing unit tests pass locally" box can be ticked.

**The fix is correct.** I reproduced #401 through `ListRulesForObject` (not just the helper) with a fake clientset: on `main` a pod in `dev` matches a binding whose `namespaceSelector` selects only `devel`; on this branch it does not. `namespaceListHasName` itself is right, including the `nil` and empty-list guards.

I found **no merge-blocking defect in the change itself**. Two things I'd still want before merge, and two follow-ups:

**Must-fix**

1. **Add a regression test at the `ListRulesForObject` level.** `TestNamespaceListHasName` only tests the new helper in isolation — it would still pass if the call site on `cache.go:122` were reverted to `strings.Contains`. The bug reported in #401 lives in the wiring, and the existing `k8sinterface.NewKubernetesApiMock()` already gives you a fake clientset whose `Namespaces().List` honours the label selector, so the test is cheap. Code inline below (verified: fails on `main`, passes here).
2. **The description undersells the error handling.** Those two `_` → `err` changes fix an unrecovered nil-pointer panic that crashes the operator, not just a missing log line. Details inline on `cache.go:100`. Please call it out in the PR body / release notes, and cover it with a test — it is the higher-severity half of this PR.

**Follow-ups (not blocking)**

3. The identical bug is still live in `node-agent` — `pkg/rulebindingmanager/cache/cache.go` has both `strings.Contains(namespaces.String(), pod.GetNamespace())` and the same two discarded `LabelSelectorAsSelector` errors (checked `v0.3.38`, the version pinned in this repo's `go.mod`, and it is still there on newer tags). The operator only imports node-agent's types and interfaces, so this PR is complete for this repo — but a user hitting #401 through the runtime (non-admission) path is not fixed until node-agent is. Worth an issue there.
4. The namespace check LISTs every matching namespace from the API server, once per rule binding, on every admission request — in the webhook latency path. A single `Namespaces().Get(object.GetNamespace())` (ideally from an informer cache) plus `nsSelector.Matches(labels.Set(ns.Labels))` is one cheap call, removes the fan-out, and removes the need for `namespaceListHasName` altogether. Detail inline on `helpers.go`.

Nits inline. Nice, tight fix — thanks for the clear comment explaining *why* `String()` must not be used.

matthyx (approved):

PR #405, opened 2026-08-21
Honour continuousScanning.matchingRules.namespaces

### LINKED ISSUE #397: continuousScanning.matchingRules.namespaces is parsed but never used - namespace scoping is silently ignored
### Description

`MatchingRules.Namespaces` is declared and unmarshalled, but `LoadGVRs` only ever reads `.APIResources`. The namespace list is dead config - continuous scanning watches every GVR cluster-wide regardless of what the user configures.

`continuousscanning/loader.go:21-25`:

```go
type MatchingRules struct {
	APIResources []APIResourceMatch `json:"match"`
	Namespaces   []string           `json:"namespaces"`   // <-- never read
}
```

`continuousscanning/loader.go:65-77`:

```go
func (l *targetLoader) LoadGVRs(ctx context.Context) []schema.GroupVersionResource {
	gvrs := []schema.GroupVersionResource{}
	rules, _ := l.fetcher.Fetch(ctx)
	apiResourceMatches := rules.APIResources   // .Namespaces is dropped on the floor
	for idx := range apiResourceMatches {
		gvrs = append(gvrs, matchRuleToGVR(apiResourceMatches[idx])...)
	}
	return gvrs
}
```

`LoadGVRs` is the only method on the `TargetLoader` interface, so the namespace list has nowhere to go. The watch pool is then built with an empty `metav1.ListOptions{}` and no namespace restriction (`continuousscanning/service.go:29-34`). The only namespace filtering that actually happens is `cfg.SkipNamespace()` in `service.go:47`, which is the unrelated global exclude list.

This matters because the Helm chart ships a namespace list by default and documents it as functional    `helm-charts/charts/kubescape-operator/values.yaml`:

```yaml
continuousScanning:
  configMapName: cs-matching-rules
  # Matching rules for the monitored resources.
  # Kubescape will watch resources of every provided GVR across the provided
  # namespaces.
  matchingRules:
    match:
      - apiGroups: ["apps"]
        apiVersions: ["v1"]
        resources: ["deployments"]
    namespaces:
      - default
```

### Reproduction

Install with the default chart values (`namespaces: [default]`) and create Deployments in a namespace other than `default`. They get scanned anyway.

Observed in the wild on a user's cluster (Slack #kubescape): with stock defaults, the operator triggered scans for Deployments in `app-workloads` and `local-path-storage`, neither of which is `default`:

```
{"msg":"fetched gvrs","gvrs":[{"Group":"apps","Version":"v1","Resource":"deployments"}]}
{"msg":"triggering scan","kind":"Deployment","name":"backend-api","namespace":"app-workloads"}
{"msg":"triggering scan","kind":"Deployment","name":"local-path-provisioner","namespace":"local-path-storage"}
```

### Expected

Either the namespace list is honoured (watches are established per-namespace, or events from non-listed namespaces are filtered before dispatch), or the field is removed from the struct and the chart so it stops implying scoping that doesn't exist.

Honouring it is the smaller surprise for users, but it is a behaviour change for anyone running the default `namespaces: [default]` today - they'd go from cluster-wide scanning to `default`- only on upgrade. Worth deciding explicitly. Happy to open a PR either way once there's a preference.

arpitjain099:
Closes #397.

`MatchingRules.Namespaces` is declared and unmarshalled (`continuousscanning/loader.go:24`) and then never read. `LoadGVRs` takes `.APIResources` and nothing else, and `LoadGVRs` was the entire `TargetLoader` interface, so the namespace list had nowhere to go. `NewDynamicWatch` then hardcoded `Namespace("")`, so continuous scanning watched every GVR across the whole cluster regardless of what was configured.

That matters because the Helm chart ships a `namespaces` list in its defaults and documents it as functional:

```yaml
continuousScanning:
  matchingRules:
    match:
      - apiGroups: ["apps"]
        apiVersions: ["v1"]
        resources: ["deployments"]
    namespaces:
      - default
```

so it reads as namespace scoping that quietly does nothing. The only namespace filtering that actually happened is `cfg.SkipNamespace` in `service.go:47`, which is the unrelated global exclude list.

The change adds `LoadNamespaces` to `TargetLoader`, threads it from `service.listen` into `NewWatchPool`, and builds one watch per GVR per namespace. Two things kept deliberately unchanged:

- Cluster-scoped resources still ignore the namespace, via the existing `k8sinterface.IsNamespaceScope` branch, since there is nothing to scope.
- An empty or absent list still means every namespace, so any rule set that omits the field behaves exactly as it does now. That is the case the `no namespaces keeps one watch per gvr` test pins.

Tests: `TestNewWatchPoolNamespaces` covers both the empty case and the fan-out (2 GVRs by 2 namespaces gives 4 watches, one per pair), `TestNewDynamicWatch` gains a case asserting the namespace reaches the recorded watch action on the fake client, and `TestTargetLoaderLoadNamespaces` covers the loader in both shapes. `assertWatchAction` now checks the namespace as well as the GVR, which is what makes the first of those meaningful.

`go build ./continuousscanning/...`, `go vet ./continuousscanning/...` and `go test ./continuousscanning/...` are clean. A whole-repo `go build ./...` does not complete on macOS: `inspektor-gadget/pkg/utils/host` is Linux-only and excluded by build constraints here. That is pre-existing and unrelated to this change, but it does mean I have not built the packages that depend on it, so CI is the real check for those.

## Summary by CodeRabbit

* **New Features**
  * Added namespace-aware continuous scanning for supported resources.
  * Scans can now target specific namespaces or retain cluster-wide coverage when none are configured.
  * Watch pools create separate watches for each configured namespace.
  * Cluster-scoped resources continue to be monitored independently of namespace settings.

### DIFF
--- a/continuousscanning/watchbuilder.go
+++ b/continuousscanning/watchbuilder.go
@@ -13,11 +13,16 @@ import (
 	"k8s.io/client-go/dynamic"
 )
 
-func NewDynamicWatch(ctx context.Context, client dynamic.Interface, gvr schema.GroupVersionResource, opts metav1.ListOptions) (watch.Interface, error) {
+// NewDynamicWatch starts a watch for the given GVR.
+//
+// namespace restricts a namespace-scoped resource to that namespace. An
+// empty namespace watches all of them. A cluster-scoped resource ignores
+// it, since there is nothing to scope.
+func NewDynamicWatch(ctx context.Context, client dynamic.Interface, gvr schema.GroupVersionResource, namespace string, opts metav1.ListOptions) (watch.Interface, error) {
 	var w watch.Interface
 	var err error
 	if k8sinterface.IsNamespaceScope(&gvr) {
-		w, err = client.Resource(gvr).Namespace("").Watch(ctx, opts)
+		w, err = client.Resource(gvr).Namespace(namespace).Watch(ctx, opts)
 	} else {
 		w, err = client.Resource(gvr).Watch(ctx, opts)
 	}
@@ -28,14 +33,16 @@ type SelfHealingWatch struct {
 	opts          metav1.ListOptions
 	client        dynamic.Interface
 	currWatch     watch.Interface
-	makeWatchFunc func(ctx context.Context, client dynamic.Interface, gvr schema.GroupVersionResource, opts metav1.ListOptions) (watch.Interface, error)
+	makeWatchFunc func(ctx context.Context, client dynamic.Interface, gvr schema.GroupVersionResource, namespace string, opts metav1.ListOptions) (watch.Interface, error)
 	gvr           schema.GroupVersionResource
+	namespace     string
 }
 
-func NewSelfHealingWatch(client dynamic.Interface, gvr schema.GroupVersionResource, opts metav1.ListOptions) *SelfHealingWatch {
+func NewSelfHealingWatch(client dynamic.Interface, gvr schema.GroupVersionResource, namespace string, opts metav1.ListOptions) *SelfHealingWatch {
 	return &SelfHealingWatch{
 		client:        client,
 		gvr:           gvr,
+		namespace:     namespace,
 		opts:          opts,
 		makeWatchFunc: NewDynamicWatch,
 	}
@@ -67,7 +74,7 @@ func (w *SelfHealingWatch) Run(ctx context.Context, readyWg *sync.WaitGroup, out
 		default:
 			gvr := helpers.String("gvr", w.gvr.String())
 			logger.L().Debug("creating watch for GVR", gvr)
-			watchFunc, err := w.makeWatchFunc(ctx, w.client, w.gvr, w.opts)
+			watchFunc, err := w.makeWatchFunc(ctx, w.client, w.gvr, w.namespace, w.opts)
 			if err != nil {
 				logger.L().Ctx(ctx).Warning(
 					"got error when creating a watch for gvr",
@@ -109,14 +116,22 @@ func (wp *WatchPool) Run(ctx context.Context, out chan<- watch.Event) {
 	logger.L().Info("Watch pool: started ok")
 }
 
-func NewWatchPool(_ context.Context, client dynamic.Interface, gvrs []schema.GroupVersionResource, opts metav1.ListOptions) (*WatchPool, error) {
-	watches := make([]*SelfHealingWatch, len(gvrs))
+// NewWatchPool builds one watch per GVR per namespace.
+//
+// An empty namespaces slice keeps the previous behaviour of a single watch
+// per GVR across the whole cluster.
+func NewWatchPool(_ context.Context, client dynamic.Interface, gvrs []schema.GroupVersionResource, namespaces []string, opts metav1.ListOptions) (*WatchPool, error) {
+	if len(namespaces) == 0 {
+		namespaces = []string{""}
+	}
+
+	watches := make([]*SelfHealingWatch, 0, len(gvrs)*len(namespaces))
 
 	for idx := range gvrs {
 		gvr := gvrs[idx]
-		selfHealingWatch := NewSelfHealingWatch(client, gvr, opts)
-
-		watches[idx] = selfHealingWatch
+		for _, namespace := range namespaces {
+			watches = append(watches, NewSelfHealingWatch(client, gvr, namespace, opts))
+		}
 	}
 
 	pool := &WatchPool{pool: watches}
--- a/continuousscanning/watchbuilder.go
+++ b/continuousscanning/watchbuilder.go
@@ -116,7 +116,11 @@ func (wp *WatchPool) Run(ctx context.Context, out chan<- watch.Event) {
 	logger.L().Info("Watch pool: started ok")
 }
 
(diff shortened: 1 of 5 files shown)


continuousscanning/watchbuilder.go
```diff
@@ -109,14 +116,22 @@ func (wp *WatchPool) Run(ctx context.Context, out chan<- watch.Event) {
...
+
+	watches := make([]*SelfHealingWatch, 0, len(gvrs)*len(namespaces))
 
 	for idx := range gvrs {
 		gvr := gvrs[idx]
-		selfHealingWatch := NewSelfHealingWatch(client, gvr, opts)
-
-		watches[idx] = selfHealingWatch
+		for _, namespace := range namespaces {
+			watches = append(watches, NewSelfHealingWatch(client, gvr, namespace, opts))
```
matthyx, line 145: **Blocker:** this fans out one watch per (GVR, namespace) pair unconditionally, including for cluster-scoped GVRs. `NewDynamicWatch`/`IsNamespaceScope` correctly ignores `namespace` for a cluster-scoped resource, but that just means the *same* cluster-wide watch gets created once per configured namespace — e.g. with `namespaces: [default, kube-system]` and a cluster-scoped resource in `APIResources` (nodes, clusterroles, etc.), you get 2 identical `SelfHealingWatch`es both streaming every event for that resource into the same channel. That's duplicate events feeding the scanner, not just wasted watches.

Fix: skip the per-namespace fan-out for cluster-scoped GVRs and create exactly one watch for them (e.g. check `k8sinterface.IsNamespaceScope(&gvr)` here, or default `namespaces` to `[""]` per-GVR instead of once globally). `TestNewWatchPoolNamespaces` only uses namespaced GVRs (`deployments`, `pods`), so it doesn't catch this — worth adding a cluster-scoped case (e.g. `clusterroles`) asserting exactly one watch regardless of how many namespaces are configured.

matthyx (changes requested):
Solid fix for the underlying issue (#397) — the loader/service/watchbuilder wiring and the empty-namespaces-means-all-namespaces fallback all look right, and the new tests cover the intended fan-out.

One blocker before merge: cluster-scoped resources get a duplicate watch per configured namespace (see inline comment on `watchbuilder.go`), since the per-namespace fan-out in `NewWatchPool` doesn't special-case `IsNamespaceScope`. That means duplicate events into the scan channel whenever a matching-rule set mixes cluster-scoped and namespace-scoped resources with a non-empty `namespaces` list — plausible given the shipped Helm defaults. CodeRabbit's automated review flagged the same thing independently.

Not requesting changes beyond that — will re-review once the fan-out is scoped to namespaced GVRs only (or namespaces default to `[""]` per-GVR rather than globally) and a cluster-scoped case is added to `TestNewWatchPoolNamespaces`.

continuousscanning/watchbuilder.go
arpitjain099: You are right, and the consequence is worse than wasted watches as you say: the duplicates all write into the same `resourceEventsCh`, so the scanner sees each event once per configured namespace.

Fixed in ab9bc9c. The namespace list is now chosen per GVR rather than once for the pool:

```go
watchNamespaces := namespaces
if !k8sinterface.IsNamespaceScope(&gvr) {
    watchNamespaces = clusterScoped
}
```

I confirmed `IsNamespaceScope` actually resolves these offline before relying on it in a test, since it would be an easy thing to assume:

```
apps/v1, Resource=deployments                  IsNamespaceScope=true
/v1, Resource=pods                             IsNamespaceScope=true
rbac.authorization.k8s.io/v1, Resource=clusterroles IsNamespaceScope=false
/v1, Resource=nodes                            IsNamespaceScope=false
```

`TestNewWatchPoolClusterScopedGVRs` covers three cases: two cluster-scoped GVRs against three namespaces, the mixed set you described, and a cluster-scoped GVR with no namespaces configured. Without the change the first two fail with exactly the duplication you predicted:

```
expected: map[string]int{"":2}
actual  : map[string]int{"default":2, "kube-system":2, "kubescape":2}

expected: map[string]int{"":1, "default":1, "kube-system":1}
```

so six watches where two were wanted. The mixed case asserts the exact namespace distribution rather than just the count, so a fix that under-fans the namespaced GVR would not pass either.

I left `TestNewWatchPoolNamespaces` alone since it still covers the namespaced fan-out on its own terms.

`go build`, `go vet` and `go test ./continuousscanning/...` are clean. As before, a whole-repo `go build ./...` does not complete on macOS because `inspektor-gadget/pkg/utils/host` is Linux-only, so CI is the check for the packages that depend on it.

matthyx (approved):
The blocker is fixed in ab9bc9c — `NewWatchPool` now picks the namespace list per GVR (`clusterScoped = [""]` for anything `!IsNamespaceScope`), so a cluster-scoped resource gets exactly one watch regardless of the configured namespaces, while namespaced GVRs still fan out correctly. `TestNewWatchPoolClusterScopedGVRs` covers the cluster-scoped-only, mixed, and no-namespaces cases, including asserting the exact namespace distribution for the mixed case so an under-fan on the namespaced side wouldn't slip through either. CI is green.

Approving.

PR #406, opened 2026-08-21
fix: log the actual kubescape response body on scan errors

magic-peach:
`readKubescapeV1ScanResponse` builds its error with `fmt.Errorf("...body: %s", resp.StatusCode, resp.Body)` when the scan endpoint returns a non-200. `resp.Body` is an `io.ReadCloser`, so `%s` never prints the actual response text — it prints Go's reflection dump of the internal `*http.body` struct instead, which is useless for figuring out why kubescape rejected the scan.

Moved the `io.ReadAll` above the status check (it already happens on the success path) and used the bytes in the error message instead. Added a test against a real `httptest` server so it actually exercises the concrete `resp.Body` type instead of a hand-built one that would mask this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for failed scan responses by including details returned by the server.
  * Ensured response-reading errors are reported clearly before validating the HTTP status.

### DIFF
--- a/mainhandler/kubescapehandlerhelper.go
+++ b/mainhandler/kubescapehandlerhelper.go
@@ -90,13 +90,14 @@ func readKubescapeV1ScanResponse(resp *http.Response) (*utilsmetav1.Response, er
 		return response, nil
 	}
 	defer resp.Body.Close()
-	if resp.StatusCode != http.StatusOK {
-		return response, fmt.Errorf("received status code '%d' from kubescape, body: %s", resp.StatusCode, resp.Body)
-	}
 
 	bodyBytes, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return response, err
+	}
+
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Confirmed the bug: `%s` on `resp.Body` (an `io.ReadCloser`) was printing Go's struct dump instead of the response text, so scan-failure logs never showed what kubescape actually said. Moving `io.ReadAll` above the status check and formatting `string(bodyBytes)` fixes this correctly — the body is still read/closed exactly once regardless of status, and the only caller (`kubescapehandler.go`) just checks the returned error, so no behavior elsewhere depends on the old ordering. The added `httptest`-backed test genuinely exercises the concrete `resp.Body` type that caused the original bug, and it responds synchronously so there's no CI-hang risk. No blockers.

PR #407, opened 2026-08-21
fix(remediation): honour selector.namespace when resolving findings-driven targets

yugal07:
## The problem

`OperatorActionSelector.Namespace` is part of the `operatorAction` contract and is documented as limiting a findings-driven selector to a single namespace:

```go
// Namespace optionally limits the selector to a single namespace.
Namespace string `json:"namespace,omitempty"`
```

`resolveSelectorTargets` never read it. It always listed cluster-wide:

```go
summaries, err := actionHandler.ksStorageClient.SpdxV1beta1().
    WorkloadConfigurationScanSummaries("").List(ctx, metav1.ListOptions{})
```

So a request to quarantine workloads failing C-0016 in `payments` would resolve to every workload failing C-0016 anywhere in the cluster, and nothing in the response or the events would indicate the scope was wider than asked for. For an action set whose whole point is to mutate running workloads, silently ignoring the field that bounds the blast radius is the wrong failure mode.

## The fix

Two parts, because there are two different namespaces in play.

The listing is now scoped server-side to `selector.Namespace`, which is both correct and cheaper than filtering client-side. An empty `Namespace` still lists across all namespaces, so cluster-wide selectors behave exactly as before.

The resolver also drops any summary whose label-derived target namespace falls outside the requested one. That namespace comes from the `kubescape.io/workload-namespace` label and is what the remediator actually acts on, so if it ever disagreed with the namespace the summary object is stored in, the server-side scoping alone would not be enough. This is defence in depth rather than a bug I observed in practice, but the cost is one comparison and the downside of getting it wrong is mutating a workload in a namespace the caller did not name.

## Context

Part of the phased `operatorAction` remediation framework:

- Proposal: kubescape/designs-and-proposals#5
- Original issue: kubescape/kubescape#1770
- Command contract, where the Namespace field is defined: armosec/armoapi-go#655
- Findings-driven targeting, which this fixes: #391
- Related work in the operator: #383 and #388

Found while wiring the CLI side of findings-driven targeting in kubescape/kubescape#3474. That PR adds `--control` and `--min-severity`, and deliberately rejects `--target-namespace` combined with a selector precisely because of this bug, since offering the flag while the operator ignores it would be worse than not offering it at all.

Once this merges and ships in a release, the CLI guard can be lifted and `--target-namespace` passed through. The guard needs to outlive this fix by a release, because a CLI that sends `selector.namespace` to an older operator would hit exactly the silent widening described above.

## Testing

Five new tests in `mainhandler/findings_test.go`:

- a namespaced selector resolves only to workloads in that namespace
- an empty namespace still resolves cluster-wide, so existing behaviour is unchanged
- namespace scoping applies to a severity-only selector, not just a control-based one
- a namespace with no matching findings resolves to nothing, which the caller already reports as an error
- a summary whose labels point outside the requested namespace is dropped

I confirmed the three scoping tests fail against `main` without the change and pass with it, so they are genuine regression tests rather than tests written to match current behaviour. Full `mainhandler` and `mainhandler/remediators` suites pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Findings-based selectors now correctly respect the requested namespace.
  * Empty namespaces continue to support cluster-wide resolution.
  * Targets outside the requested namespace are excluded from summaries.
* **Tests**
  * Added coverage for namespace filtering, cluster-wide behavior, missing findings, and mismatched target labels.

### DIFF
--- a/mainhandler/findings.go
+++ b/mainhandler/findings.go
@@ -50,7 +50,8 @@ func severityRank(severity string) (int, bool) {
 // No re-scan happens: it acts on findings already in the cluster.
 //
 // A summary matches when it fails the requested Control and/or carries a failing
-// finding at or above MinSeverity (see summaryMatchesSelector). Targets in
+// finding at or above MinSeverity (see summaryMatchesSelector), and, when the
+// selector names a Namespace, when it resolves to a workload in it. Targets in
 // excluded/protected namespaces and non-remediable kinds are dropped here so
 // they never reach a remediator, and duplicates are collapsed.
 func (actionHandler *ActionHandler) resolveSelectorTargets(ctx context.Context, selector *apis.OperatorActionSelector) ([]remediators.Target, error) {
@@ -73,8 +74,11 @@ func (actionHandler *ActionHandler) resolveSelectorTargets(ctx context.Context, 
 		return nil, fmt.Errorf("operatorAction: findings-driven targeting requires the storage client, which is not configured")
 	}
 
-	// namespace "" lists configuration-scan summaries across all namespaces.
-	summaries, err := actionHandler.ksStorageClient.SpdxV1beta1().WorkloadConfigurationScanSummaries("").List(ctx, metav1.ListOptions{})
+	// Scope the listing to the requested namespace. An empty Namespace lists
+	// across all of them, which stays the default for a cluster-wide selector.
+	// Without this the selector always resolved cluster-wide, so a caller asking
+	// for one namespace silently got every matching workload in the cluster.
+	summaries, err := actionHandler.ksStorageClient.SpdxV1beta1().WorkloadConfigurationScanSummaries(selector.Namespace).List(ctx, metav1.ListOptions{})
 	if err != nil {
 		return nil, fmt.Errorf("operatorAction: failed to list configuration scan summaries: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


matthyx (approved):
Reviewed the diff, the resolver logic, and the new tests.

- Listing is now scoped server-side via `WorkloadConfigurationScanSummaries(selector.Namespace)`, and an empty namespace correctly falls back to cluster-wide, matching the pre-existing behavior for unscoped selectors.
- The added defense-in-depth check (`target.Namespace != selector.Namespace`) is sound: it guards against a summary's label-derived namespace disagreeing with the object's own namespace, which is the namespace the remediator actually acts on.
- The five new tests in `mainhandler/findings_test.go` cover the namespaced case, the cluster-wide/empty-namespace case, namespace scoping combined with `MinSeverity`, a namespace with no matches, and the mislabelled-summary defense-in-depth case. Good regression coverage, and the description notes the scoping tests were confirmed to fail on `main` without the fix.
- All CI checks (build, tests, CodeQL, GitGuardian, DCO, FOSSA) are green. CodeRabbit's automated review found no actionable comments — only a generic "docstring coverage" heuristic warning, which isn't a real blocker for this package's conventions.

No blockers. This is a correct, well-tested, appropriately-scoped fix. Approving.

PR #411, opened 2026-09-04
fix(restapi): allowlist commandName on the unauthenticated triggerAction endpoint

matthyx:
## Overview

`/v1/triggerAction` has no caller authentication, and its Service carries no NetworkPolicy — any pod on the cluster network can call it. Its shared dispatcher previously accepted every `CommandName` the operator recognizes, including `operatorAction` (the command type behind `annotate`/`quarantine`/`revert`/`patch`). That meant an anonymous caller with zero credentials could direct the operator's cluster-wide remediation RBAC at arbitrary workloads.

This closes that specific gap by allowlisting exactly the `CommandName`s the endpoint's real callers send, and rejecting everything else — including `operatorAction`.

## Context

Found while validating a security finding on #410 (the new generic `patch` remediation action). I verified this live on `armo-dev-stage`: an anonymous, tokenless POST from a throwaway in-cluster pod to `operator:4002/v1/triggerAction` returned `200`/`ok` with no auth challenge at any layer.

I then confirmed exactly who the legitimate callers are and what they send, live:

| CronJob | `commandName` sent |
|---|---|
| `kubescape-scheduler` | `kubescapeScan` |
| `kubevuln-scheduler` | `scan` |
| registry-scan CronJob (created dynamically) | `scanRegistryV2` |

None of them ever send `operatorAction`. The backend's actual command channel is unrelated to this endpoint entirely — it goes through the `synchronizer` component's own authenticated outbound connection, which creates `OperatorCommand` CRs gated by real Kubernetes RBAC (`ClusterRole/synchronizer`). `/v1/triggerAction` is a separate, secondary HTTP door into the same dispatch code, with none of that path's guardrails.

## What's in this PR

- `restapihandler/triggeraction.go`: `triggerActionAllowedCommands` allowlist, checked per-command before dispatch (same rejection pattern already used for an empty `CommandName` — logged, recorded on status as a no-op, skipped; HTTP response stays `200`/`ok`, unchanged).
- `restapihandler/triggeraction_test.go`: covers an allowed command being dispatched, `operatorAction` (including the exact shape a `patch` action would need) being rejected, an unknown command being rejected, a mixed batch dispatching only the allowed entry, and the HTTP-level response shape for a rejected command.
- `docs/features/trigger-action-command-allowlist.md`: what changed and why.

## Follow-up

This is a stopgap, not a full fix — the endpoint is still unauthenticated for the commands it does allow. The real fix is to stop using an unauthenticated HTTP callback here at all and have these CronJobs create `OperatorCommand` CRs directly (the same authenticated/RBAC-gated mechanism `synchronizer` already uses), removing this HTTP path entirely. That requires Helm chart changes, so I'm opening a separate tracking issue for it rather than bundling it here.

## How to test

```
go build ./...
go vet ./...
go test ./restapihandler/... -v
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018MmNHibtv3BGpGLqzWYgwQ

AI-skills: oh-my-claudecode:cancel | cmds: /oh-my-claudecode:autopilot

## Summary by CodeRabbit

* **Security**
  * Restricted trigger-action requests to approved scan commands: `kubescapeScan`, `scan`, and `scanRegistryV2`.
  * Operator actions and unrecognized or empty commands are rejected before execution.
  * Rejected commands are recorded as errors while the endpoint continues returning a successful HTTP response.

* **Documentation**
  * Added guidance describing the trigger-action command allowlist and rejection behavior.

* **Tests**
  * Added coverage for approved, rejected, mixed, and empty command requests.

### DIFF
--- a/docs/features/trigger-action-command-allowlist.md
+++ b/docs/features/trigger-action-command-allowlist.md
@@ -1,0 +1,52 @@
+# triggerAction command allowlist
+
+## Overview
+
+`/v1/triggerAction` (`restapihandler/triggeraction.go`, port `4002`, named
+`trigger-port`) is an unauthenticated HTTP endpoint: it has no caller
+identity check, and its Service has no NetworkPolicy restricting who can
+reach it. It now only dispatches an explicit allowlist of `CommandName`
+values instead of every command the operator recognizes.
+
+## Why
+
+The endpoint's only legitimate callers are the operator's own
+scan-scheduling CronJobs, run in-cluster with a fixed, hardcoded request
+body:
+
+| CronJob | `commandName` sent |
+|---|---|
+| `kubescape-scheduler` | `kubescapeScan` (`apis.TypeRunKubescape`) |
+| `kubevuln-scheduler` | `scan` (`apis.TypeScanImages`) |
+| registry-scan CronJob (created dynamically when registry scanning is configured — see `watcher/registryhandler.go`) | `scanRegistryV2` (`apis.TypeScanRegistryV2`) |
+
+Before this change, the endpoint's shared dispatcher (`handleRequest` in
+`mainhandler/handlerequests.go`) accepted every `CommandName` it recognizes,
+including `operatorAction` — the command type that drives remediation
+(`annotate`/`quarantine`/`revert`/`patch`). Because nothing gates who can
+reach port 4002, any pod on the cluster network could send an `operatorAction`
+command and have the operator apply it with its own cluster-wide remediation
+RBAC — a materially different risk than the legitimate callers' fixed,
+read-only-triggering payloads.
+
+This is a stopgap: the real fix is to stop using an unauthenticated HTTP
+callback for this at all, and have these CronJobs create `OperatorCommand`
+custom resources directly instead (the same, RBAC-gated mechanism the
+`synchronizer` component already uses for backend-originated commands). That
+migration is tracked separately since it requires Helm chart changes.
+
+## What changed
+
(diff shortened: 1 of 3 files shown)


matthyx:
Closing this in favor of a narrower fix on #410.

While digging into how commands actually reach `/v1/triggerAction`, I found [`designs-and-proposals/cli-cluster-operations.md`](https://github.com/kubescape/designs-and-proposals/blob/50e4e2b15857595469bf82cc0246dea820a6020a/proposals/cli-cluster-operations.md), which documents this endpoint accepting `operatorAction` as the **intended transport** for the `kubescape` CLI's `operator remediate annotate|quarantine|revert` subcommand (reached via `kubectl port-forward`, itself RBAC-gated). The allowlist here excludes `operatorAction` entirely, which would break that real, shipped feature — it has no other transport. My mistake for not checking the CLI/design doc before opening this.

The endpoint's underlying gap (no application-level auth, no `NetworkPolicy` on its Service) is still real and still worth closing — but that needs a `NetworkPolicy`-level fix in `kubescape/helm-charts` (since `kubectl port-forward` traffic doesn't cross the pod network a `NetworkPolicy` governs, it wouldn't break the CLI), not a `commandName` allowlist here. The specific risk that motivated this PR — the new `patch` action being reachable through this unauthenticated path — is now addressed more narrowly in #410 instead, by gating `patch` on the `OperatorCommand` CRD delivery path (which `annotate`/`quarantine`/`revert` don't need, since they're the actions the CLI/design doc actually covers).

PR #413, opened 2026-09-04
fix(continuousscanning): fail startup on invalid matching rules config

### LINKED ISSUE #399: Operator panics with a nil pointer dereference when cs-matching-rules contains null or malformed JSON
# Description

If the `cs-matching-rules` configmap contains `null` or malformed JSON, the operator panics with a
nil pointer dereference on startup instead of reporting a configuration error. Because the panic
happens in an unrecovered goroutine, it takes the whole operator process down - including admission,
vulnerability scan triggering, and the node-agent autoscaler.

The cause is in `continuousscanning/loader.go`:

```go
func (l *targetLoader) LoadGVRs(ctx context.Context) []schema.GroupVersionResource {
	gvrs := []schema.GroupVersionResource{}

	rules, _ := l.fetcher.Fetch(ctx)      // error dropped

	apiResourceMatches := rules.APIResources    // nil deref
	...
}
```

`parseMatchingRules` unmarshals into a `*MatchingRules`, so `rules` comes back nil in two cases:

- **malformed or empty JSON** - `Unmarshal` returns an error, which is discarded on the line above,
  and the nil pointer is then dereferenced.
- **the literal `null`** - `Unmarshal` returns **no error at all** and leaves the pointer nil. Adding
  error handling alone would not catch this one; it needs an explicit nil check.

Note that a *missing* file is already handled correctly (`mainhandler/handlerequests.go:122-125`
returns the `os.Open` error and `main.go` turns it into a clean fatal). It's specifically the
"file exists, content is bad" case that panics.

# Environment
OS: `Linux` 
Version: `operator main @ a1299a5` 

# Steps To Reproduce

1. The chart renders the configmap with `{{ mustToJson .Values.continuousScanning.matchingRules }}`,
   so setting that value to null puts the literal string `null` into the configmap:

   ```console
   $ helm template ks charts/kubescape-operator \
       --set continuousScanning.matchingRules=null \
       -s templates/configs/matchingRules-configmap.yaml
   data:
     matchingRules.json: |
       null
   ```

2. Install with that value and continuous scanning enabled:

   ```console
   $ helm install kubescape kubescape/kubescape-operator \
       --set capabilities.continuousScan=enable \
       --set continuousScanning.matchingRules=null
   ```

3. The operator pod crashes. A hand-edited or truncated `cs-matching-rules` configmap does the same.

Or reproduce it directly as a unit test in `continuousscanning`:

```go
func TestLoadGVRs_NullJSON(t *testing.T) {
	loader := NewTargetLoader(NewFileFetcher(strings.NewReader("null")))
	defer func() {
		if r := recover(); r != nil {
			t.Fatalf("LoadGVRs panicked: %v", r)
		}
	}()
	loader.LoadGVRs(context.Background())
}
```

# Expected behavior

A bad matching-rules configmap should produce a clear configuration error, not a panic. Either fail
startup with a message naming the file and the parse problem, or start with continuous scanning
disabled and log loudly - but not crash.

One thing worth preserving either way: `{"match":[]}` is a legitimate "watch nothing" configuration
and should keep working normally. Only unparseable or `null` input should be treated as an error.

# Actual Behavior

The operator panics. Running the test above against `"null"`, `"{not json"` and `""`, plus two
controls for valid input:

```
--- FAIL: TestLoadGVRs_NullJSON        LoadGVRs panicked: nil pointer dereference
--- FAIL: TestLoadGVRs_MalformedJSON   LoadGVRs panicked: nil pointer dereference
--- FAIL: TestLoadGVRs_EmptyFile       LoadGVRs panicked: nil pointer dereference
--- PASS: TestLoadGVRs_ExplicitlyEmpty     0 gvrs, no panic
--- PASS: TestLoadGVRs_ChartDefault        [apps/v1, Resource=deployments]
```

# Additional context

The reason the error is dropped in the first place is that `LoadGVRs` has no error return, so there
is nowhere to put it. I'd suggest giving it one:

```go
func (l *targetLoader) LoadGVRs(ctx context.Context) ([]schema.GroupVersionResource, error) {
	rules, err := l.fetcher.Fetch(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to fetch matching rules: %w", err)
	}
	if rules == nil {
		return nil, errors.New("matching rules are null")
	}
	...
}
```

That requires `listen` and `Launch` to return errors too - but both currently return a
`<-chan armoapi.Command` that **nothing ever sends on** (`service.go:27`/`:59` and `:90`/`:95`), and
all three callers discard the return value. So it's closer to swapping two dead return values for a
live one than to adding plumbing. Giving `listen` an error return would also pick up the second
dropped error a couple of lines down (`wp, _ := NewWatchPool(...)`, `service.go:34`) - harmless today
because `NewWatchPool` never returns a non-nil error, but the same trap is sitting there.

The error would then land in `SetupContinuousScanning`, which already returns an error to a caller
that already handles it. That also moves the "fatal, or start with continuous scanning disabled?"
decision into `main.go` as a one-line choice, alongside the same decisions already being made for the
HTTP exporter and the node-agent autoscaler, instead of hardcoding it in the loader. I'd lean fatal
to match the storage-client fatal already in `SetupContinuousScanning`, but either is defensible.

Two small things in the same code path:

- `rulesReader` is never closed (`mainhandler/handlerequests.go:122`).
- `main.go:165-172` logs `"set up cont scanning service"` *before* checking the returned error, so a
  failure prints a success line and then fatals.

ANAMASGARD:
Harden parseMatchingRules and consolidate TargetLoader to a single Load() that fetches once. Propagate errors through listen/Launch when cs-matching-rules contains null, malformed, or empty JSON instead of panicking. Include the matching-rules filename in the setup error. Close the rules file after open and log success only after SetupContinuousScanning succeeds.

Fixes #399

## Overview
This PR fixes #399.

Previously, if the `cs-matching-rules` ConfigMap contained `null`, malformed JSON, or empty content, `LoadGVRs()` discarded the fetch error and dereferenced a nil `*MatchingRules`, panicking in a background goroutine and taking down the whole operator.

Now bad matching-rules input returns a clear configuration error (including the filename) and fails continuous-scanning setup fatally, instead of panicking. Valid `{"match":[]}` still works as a legitimate "watch nothing" config. `TargetLoader` is consolidated to a single `Load()` so the one-shot file reader is only fetched once.

## How to Test

```bash
go test ./continuousscanning/... -count=1 -v
go test -race ./continuousscanning/... -count=1
go test ./... -count=1
go build ./...
```

Confirm these cases:
- `"null"`, whitespace+`null`, malformed JSON, empty/whitespace-only → error, no panic
- `{"match":[]}` → success, 0 GVRs
- chart-default config → expected GVRs + namespaces
- fetcher `(nil, nil)` / fetch error → propagated error
- `Load()` fetches exactly once
- `Launch()` with invalid config → error, no panic (`TestLaunch_InvalidMatchingRules`)

## Related issues/PRs

* Fixes #399

**Please open the PR against the `dev` branch (Unless the PR contains only documentation changes)**
```

Open against **`main`**, and keep the DCO box checked only if the commit was created with `git commit -s`.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

- **Bug Fixes**
  - Continuous scanning now reports setup and launch failures instead of silently continuing.
  - Invalid, empty, or null matching-rules files now produce clear errors rather than unexpected failures.
  - Configuration loading errors are propagated and associated with the relevant rules file.
  - Readiness is reported only after continuous scanning setup completes successfully.
  - Matching resources and namespaces are loaded consistently in a single operation.
  - Configuration data is fetched once per loading operation for more consistent results.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/continuousscanning/loader.go
+++ b/continuousscanning/loader.go
@@ -1,8 +1,11 @@
 package continuousscanning
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
+	"fmt"
 	"io"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -34,9 +37,11 @@ type targetLoader struct {
 	fetcher MatchingRuleFetcher
 }
 
+// TargetLoader loads matching-rule targets used to establish watches.
 type TargetLoader interface {
-	LoadGVRs(ctx context.Context) []schema.GroupVersionResource
-	LoadNamespaces(ctx context.Context) []string
+	// Load returns the GVRs to watch and the namespaces to restrict them to.
+	// An empty namespaces slice means every namespace.
+	Load(ctx context.Context) ([]schema.GroupVersionResource, []string, error)
 }
 
 // NewTargetLoader returns a new Target Loader
@@ -62,32 +67,22 @@ func matchRuleToGVR(apiMatch APIResourceMatch) []schema.GroupVersionResource {
 	return gvrs
 }
 
-// LoadGVRs loads GroupVersionResource definitions
-func (l *targetLoader) LoadGVRs(ctx context.Context) []schema.GroupVersionResource {
-	gvrs := []schema.GroupVersionResource{}
-
-	rules, _ := l.fetcher.Fetch(ctx)
-
-	apiResourceMatches := rules.APIResources
-	for idx := range apiResourceMatches {
-		ruleGvrs := matchRuleToGVR(apiResourceMatches[idx])
-		gvrs = append(gvrs, ruleGvrs...)
-	}
-
-	return gvrs
-}
-
-// LoadNamespaces loads the namespaces the matching rules restrict watches to.
-//
-// An empty result means every namespace, which is what a rule set that
-// omits the field asks for.
-func (l *targetLoader) LoadNamespaces(ctx context.Context) []string {
+// Load fetches matching rules once and returns the GVRs and namespaces to watch.
+func (l *targetLoader) Load(ctx context.Context) ([]schema.GroupVersionResource, []string, error) {
 	rules, err := l.fetcher.Fetch(ctx)
-	if err != nil || rules == nil {
-		return nil
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to fetch matching rules: %w", err)
+	}
+	if rules == nil {
+		return nil, nil, errors.New("matching rules are null")
 	}
 
-	return rules.Namespaces
+	gvrs := []schema.GroupVersionResource{}
+	for idx := range rules.APIResources {
+		gvrs = append(gvrs, matchRuleToGVR(rules.APIResources[idx])...)
+	}
+
+	return gvrs, rules.Namespaces, nil
 }
 
 type fileFetcher struct {
@@ -110,7 +105,17 @@ func parseMatchingRules(r io.Reader) (*MatchingRules, error) {
 		return nil, err
 	}
 
-	var matches *MatchingRules
-	err = json.Unmarshal(data, &matches)
-	return matches, err
+	trimmed := bytes.TrimSpace(data)
+	if len(trimmed) == 0 {
+		return nil, errors.New("matching rules are empty")
+	}
+	if bytes.Equal(trimmed, []byte("null")) {
+		return nil, errors.New("matching rules are null")
+	}
+
+	var matches MatchingRules
+	if err := json.Unmarshal(trimmed, &matches); err != nil {
+		return nil, fmt.Errorf("failed to parse matching rules: %w", err)
+	}
+	return &matches, nil
 }
--- a/continuousscanning/loader_test.go
+++ b/continuousscanning/loader_test.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"errors"
 	"io"
+	"strings"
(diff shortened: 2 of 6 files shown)


matthyx (changes requested):
Blocking on the unresolved readiness-ordering issue in `main.go`. This change makes invalid continuous-scanning configuration a startup-fatal error, but `isReadinessReady` is still set before `SetupContinuousScanning` runs asynchronously. With continuous scanning enabled, complete that setup successfully before publishing readiness so a pod with invalid rules cannot become ready while it is terminating. A focused regression test would be helpful if practical.

The implementation otherwise looks sound, and `go test ./... -count=1` passes locally.

ANAMASGARD:
@matthyx  I have moved isReadinessReady = true to after synchronous SetupContinuousScanning when continuous scanning is enabled. Invalid matching rules now fail before readiness is published. Non-CS behavior unchanged.

matthyx (approved):
Rechecked at 8d106e7. The readiness-ordering blocker is resolved: continuous-scanning setup now completes successfully before readiness is published. The full test suite passes locally with go test ./... -count=1.

PR #415, opened 2026-09-14
test: validate Agent Runtime framework scan requests

dakshhhhh16:
## Overview

This adds coverage for requesting and scheduling an AgentRuntimeHardening scan through the operator. I wanted to check that the framework selected by the user reaches the scanner and survives being saved for a later scheduled run.

The test covers an explicit framework selection and a framework inherited from the operator defaults. It calls the scan handler against a local HTTP server, then calls the scheduler with a fake Kubernetes client. It reads back the created CronJob and its request ConfigMap, decodes the stored command, and checks that replaying it preserves the request even when the defaults have changed.

The assertions cover framework selection, included and excluded namespaces, exceptions, and explicit host-scanner, keep-local, and cached-artifact settings. They also check the schedule, framework annotation, and ConfigMap volume reference.

The existing forwarding paths passed these checks, so no production change was needed. The request API does not carry inline control inputs. This PR covers the options that it supports; scanner-side control-input loading remains outside this test.

## How to test

```sh
go test ./mainhandler -run TestAgentRuntime -count=1
go test ./mainhandler -count=1
```

Both passed locally. The test uses a local HTTP server and fake Kubernetes client.

## Summary by CodeRabbit

- **Tests**
  - Added integration coverage confirming agent-runtime scan requests preserve their configuration during immediate execution and scheduled CronJob replay.
  - Verified behavior for default and explicitly selected frameworks.
  - Confirmed preservation of namespaces, scanner options, cached-artifact settings, exceptions, scheduling metadata, and serialized command details.


matthyx (approved):
Reviewed head `6a9c2953af74552352b68e92093ed517d207b8a1` against `main` at `0ff3fa57cf7d50ec7720bb4931e29d68f4a71d0d`.

This is useful coverage for the request behavior introduced by merged #398: both explicit `AgentRuntimeHardening` selection and configured-default selection reach the direct scan request and survive scheduled ConfigMap serialization even if defaults later change. The test also pins supported namespace filters, exceptions, and boolean scan options. #398's three earlier blockers were fixed and explicitly re-verified before merge; none recur here. Searches using `AgentRuntimeHardening`, agent-runtime, scheduled-scan, `defaultFrameworks`, `targetNames`, and `targetType` found no duplicate or superseding operator change. Merged #392 touches exception-triggered rescans but not this request-persistence path.

No blocking findings. The existing `NewSimpleClientset` deprecation note is nonblocking test maintenance.

Validation: focused test passed in a credential-free container with the repository mounted read-only; independent review also passed the full `mainhandler` package and `go vet ./mainhandler`; `gofmt` and `git diff --check` are clean. Current visible GitHub checks pass, though the Go test workflow is not present in this PR's check rollup. Coverage is intentionally bounded to operator forwarding and stored-command reconstruction; it does not execute a deployed CronJob or verify scanner-side framework/control availability.

Verdict: approve.

matthyx:
I'm opening a PR to fix the race issue with SafeMap (not related to this change)

PR #419, opened 2026-09-19
fix(watcher): skip scans for orphaned ContainerProfiles

### LINKED ISSUE #417: ContainerProfileWatch's startup relist re-queues scans for orphaned ReplicaSet generations, corrupting VulnerabilityManifestSummary on every operator restart
## Summary

`ContainerProfileWatch` re-lists and re-queues a scan for *every* `ContainerProfile` object in cluster storage on every operator process start (`watcher/containerprofilewatcher.go:50-62`) -- including profiles belonging to ReplicaSet generations that no longer exist. The default config leaves nothing to filter these out (`SkipProfilesWithoutInstances` defaults to `false` and isn't exposed in the Helm chart's `values.yaml`), and even the fallback pod lookup doesn't skip on failure (`getPod()` at line 139 just logs and continues if the pod is gone). Because the resulting `VulnerabilityManifestSummary` is keyed only by `<kind>-<name>-<container>` (workload identity, not ReplicaSet), and the summary write path has no freshness/generation check, a scan of a months-old dead ReplicaSet's profile can silently overwrite the summary for the image that's actually running today.

## Evidence (live cluster, kubescape-operator chart 1.40.4, kubevuln v0.3.441)

`ContainerProfile` objects for one StatefulSet's container, none garbage-collected across 6 ReplicaSet generations spanning 3 months:

```
statefulset-alertmanager-kube-prometheus-stack-alertmanager-85fd7f467f-alertmanager-...   2026-06-15
statefulset-alertmanager-kube-prometheus-stack-alertmanager-5885f87586-alertmanager-...   2026-06-24  <- corrupted the summary
statefulset-alertmanager-kube-prometheus-stack-alertmanager-855c768ffd-alertmanager-...   2026-06-29
statefulset-alertmanager-kube-prometheus-stack-alertmanager-57548f9ff4-alertmanager-...   2026-07-15
statefulset-alertmanager-kube-prometheus-stack-alertmanager-77cfc9576-alertmanager-...    2026-07-15
statefulset-alertmanager-kube-prometheus-stack-alertmanager-5b6df87455-alertmanager-...   2026-08-31
statefulset-alertmanager-kube-prometheus-stack-alertmanager-9c6996f9f-alertmanager-...    2026-09-15  <- actually running
```

Operator log, right after a pod restart (triggered here by node churn during an EKS AMI rotation, but any operator restart reproduces it):

```
"scanning container profile","wlid":"wlid://cluster-.../namespace-monitoring/statefulset-alertmanager-kube-prometheus-stack-alertmanager",
  "name":"statefulset-alertmanager-kube-prometheus-stack-alertmanager-5885f87586-config-reloader-7c63-5900"
```

`5885f87586` is the ReplicaSet hash replaced by five subsequent rollouts. After this scan ran, `VulnerabilityManifestSummary/statefulset-alertmanager-kube-prometheus-stack-alertmanager-alertmanager` reported `alertmanager:v0.32.2` -- several releases behind what was actually deployed (`v0.34.0`) -- and `vulnerabilitiesRef.relevant` pointed at the `5885f87586`-generation `VulnerabilityManifest`, not the current one. Deleting the summary and forcing a fresh scan while the operator was quiescent produced the correct result (`v0.34.0`, matching severities); the very next operator restart reintroduced the same corruption via the same mechanism.

## Why this is reachable, not hypothetical

Any event that restarts the operator pod triggers it: a node drain/replacement (confirmed reproduction), a chart upgrade, an OOM, a routine eviction. `ContainerProfileWatch`'s startup list (`containerprofilewatcher.go:50-62`) has no bound on age or on ReplicaSet liveness, and nothing elsewhere garbage-collects `ContainerProfile` objects when their owning ReplicaSet is deleted -- we found 6 generations going back 3 months for a single container.

## Suggested fixes (either resolves it; both together is more robust)

1. **Garbage-collect `ContainerProfile` objects when their owning ReplicaSet no longer exists** (owner references + GC, or a TTL/reconciler pass), so the startup relist has nothing stale to re-queue.
2. **Default `SkipProfilesWithoutInstances` to `true`** (or expose and default it in the Helm chart), and make `getPod()`'s failure path (`containerprofilewatcher.go:139-144`) skip dispatching the scan command rather than logging and continuing.
3. **Add a freshness/generation guard to the `VulnerabilityManifestSummary` write path**, so a scan result for a non-current ReplicaSet generation can't overwrite a summary that already reflects the current one -- the same class of fix already applied to `GetSBOM`'s one-directional version check in kubevuln#768 (a comparison that only guards one direction, so an out-of-order write silently wins).

## Environment

- kubescape-operator chart 1.40.4
- kubevuln v0.3.441 (confirmed current release at time of testing, not an outdated-pin issue)
- Reproduced across an EKS 1.34 and an EKS 1.35 cluster

ANAMASGARD:
## Overview

This PR fixes #417.

On startup, the operator relists stored ContainerProfiles. Previously, it could dispatch scans for profiles from deleted ReplicaSet generations, allowing an old image scan to reach kubevuln.

The operator now skips profiles without a live instance by default. ReplicaSet profiles require a Pod owned by that exact ReplicaSet; Pod profiles require their named Pod to exist. A failed Pod lookup no longer dispatches a scan. An explicit `skipProfilesWithoutInstances: false` remains supported.

## How to Test

- Run the focused configuration and watcher tests for the default, override, and orphaned-profile relist cases.
- Run `go build ./...`, `go vet ./...`, and `git diff --check`.

These checks passed locally. The full `go test ./...` run was blocked by unrelated tests requiring localhost sockets and Docker access in the test environment.

## Related issues/PRs

- Resolved #417

- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

* **New Features**
  * Profiles without matching instances are now skipped by default.
  * Profile matching supports complete workload selectors and verifies workload ownership and generation.

* **Bug Fixes**
  * Prevented scans for profiles linked to missing, deleted, or mismatched instances.
  * Profiles referencing unavailable pods are safely ignored.

* **Configuration**
  * The default behavior can be disabled through the `skipProfilesWithoutInstances` setting.


watcher/containerprofilewatcher.go
```diff
@@ -215,9 +222,23 @@ func (wh *WatchHandler) hasMatchingPod(labels map[string]string) bool {
 		logger.L().Debug("hasMatchingPod - failed to list pods matching selector", helpers.String("selector", labelsStr.String()), helpers.String("namespace", namespace), helpers.Error(err))
 		return false
 	}
-	if len(podList.Items) > 0 {
-		return true
+	for _, pod := range podList.Items {
+		if !podIsActive(&pod) {
+			continue
+		}
+		if labels[helpersv1.RelatedKindMetadataKey] != "ReplicaSet" {
```
matthyx: **Blocker:** only `ReplicaSet` profiles get the per-instance check. For `StatefulSet`/`DaemonSet` profiles (the exact case in #417 — `5885f87586` there is a `controller-revision-hash`), the labels are `workload-kind=StatefulSet`, `workload-name=<sts>` with an empty `instance-template-hash` (see `k8s-interface` `containerinstance.InstanceID.GetLabels()`; the revision lives only in `AlternateName`, i.e. the instance-ID string / profile name). This branch therefore returns `true` for every stale revision as long as one pod of the StatefulSet is running, so the restart relist still dispatches the old scans.

A kind-agnostic alternative: generate instance IDs for each candidate pod via `instanceidhandlerv1.GenerateInstanceIDFromRuntimeObj(pod, wh.cfg.ExcludeJsonPaths())` (as in `podwatcher.go:72`) and match `GetStringFormatted()` against the profile's `kubescape.io/instance-id` annotation.

watcher/containerprofilewatcher.go
```diff
@@ -215,9 +222,23 @@ func (wh *WatchHandler) hasMatchingPod(labels map[string]string) bool {
...
+				return true
+			}
+		}
 	}
 	// no matching pods found
 	return false
 }
+
+func podIsActive(pod *corev1.Pod) bool {
+	return pod != nil && pod.DeletionTimestamp == nil && pod.Status.Phase != corev1.PodSucceeded && pod.Status.Phase != corev1.PodFailed
```
matthyx: **Blocker:** with `skipProfilesWithoutInstances` now defaulting to `true`, excluding `Succeeded`/`Failed` pods means Job/CronJob (and bare Pod) profiles are never scanned: node-agent completes the profile when the container exits, so by the time the event arrives the pod is already `Succeeded`. On `main` (default `false`) these were scanned. Suggest keeping only the `DeletionTimestamp == nil` check here.

matthyx (changes requested):
Thanks for picking up #417 — flipping `skipProfilesWithoutInstances` to `true` and handling `kind=Pod` profiles directly in `hasMatchingPod` (previously they always fell through to a `spec.selector.matchLabels` lookup and were skipped) are both good. I built, vetted and ran `go test ./config/ ./watcher/` on `f636638` (all pass), but I found two blockers when checking the change against how instance IDs / profile labels are actually produced by `k8s-interface`.

**1. The reported scenario (StatefulSet revisions) is not covered.** The evidence in #417 is a StatefulSet (`statefulset-alertmanager-...-5885f87586-...`); that suffix is the `controller-revision-hash`, not a ReplicaSet. For StatefulSet/DaemonSet pods, `k8s-interface` puts the revision only in the instance-ID *name* (`AlternateName`) — the profile labels carry `workload-kind=StatefulSet`, `workload-name=<sts name>` and an empty `instance-template-hash`. The new owner-UID check only runs for `workload-kind=ReplicaSet`, so a stale-revision StatefulSet profile still returns `true` as long as *any* pod of the StatefulSet is running. I verified this with a scratch test on this branch (StatefulSet with one current pod, profile for an old revision → `hasMatchingPod == true`). The ReplicaSet path, on the other hand, was already handled on `main`: old ReplicaSets are either gone (`Get` fails) or scaled to 0 with a `pod-template-hash` selector, so the added UID check is defensive rather than the fix.

Suggested remedy (kind-agnostic): for each candidate pod, compute its instance IDs with `instanceidhandlerv1.GenerateInstanceIDFromRuntimeObj(pod, wh.cfg.ExcludeJsonPaths())` (already used in `watcher/podwatcher.go:72`) and compare `GetStringFormatted()` with the profile's `kubescape.io/instance-id` annotation (or compare slugs with the profile name). That matches exactly the instance the profile was learned from, for Deployments, StatefulSets, DaemonSets, Jobs and bare Pods alike, and makes the ReplicaSet-specific branch unnecessary.

**2. Regression for Job/CronJob (and bare Pod) profiles.** `podIsActive` rejects `Succeeded`/`Failed` pods, and the flag is now on by default. node-agent finalizes a ContainerProfile when the container exits, so for Jobs/CronJobs the profile arrives *after* the pod is `Succeeded` — with this PR those profiles are never dispatched (scratch test: completed Job pod → `hasMatchingPod == false`; on `main` with the default config they were scanned). This applies to live watch events too, not just the startup relist. Please drop the phase check (keeping `DeletionTimestamp == nil` is fine) or scope it to workload kinds that are expected to keep running.

Non-blocking:
- `kind=Pod` profiles now `Get` the pod twice (in `hasMatchingPod` and again in `getPod`); consider passing the pod through.
- The `wrongOwner` fixture (an old ReplicaSet whose selector lacks `pod-template-hash`) doesn't occur with Deployment-managed ReplicaSets; fine as a defensive test but it isn't the #417 case.
- The default flip isn't exposed in `kubescape/helm-charts`; a follow-up there would let users opt out without a custom configmap.

Related history: the flag and `hasMatchingPod` were introduced in #332 (Sep 2025) as opt-in; #362 skips TS profiles. No prior PR attempted #417.

Reviewed commit: `f636638` against `main` (`616d85b`).

matthyx:
Re-reviewed `1ed9313`: the switch to the full `spec.selector` via `metav1.LabelSelectorAsSelector` is correct and resolves CodeRabbit's `matchExpressions` finding — `go build`, `go vet` and `go test ./config/ ./watcher/` pass locally.

The two blockers from my review are unchanged in this commit (`hasMatchingPod` still only checks instance identity for `workload-kind=ReplicaSet`, so the StatefulSet case from #417 is still dispatched; `podIsActive` still rejects `Succeeded`/`Failed` pods, so Job/CronJob profiles are skipped by default). Keeping "changes requested" until those threads are addressed.

watcher/containerprofilewatcher.go
```diff
@@ -189,35 +198,66 @@ func (wh *WatchHandler) hasMatchingPod(labels map[string]string) bool {
...
+		return false
 	}
-	if labelsStr.Len() == 0 {
+	labelSelector, err := metav1.LabelSelectorAsSelector(&selector)
+	if err != nil || labelSelector.Empty() {
 		logger.L().Debug("hasMatchingPod - empty pod selector from workload", helpers.String("gvr", gvr.String()), helpers.String("namespace", namespace), helpers.String("name", name))
 		return false
 	}
+	var statefulSetRevision string
+	if labels[helpersv1.RelatedKindMetadataKey] == "StatefulSet" {
```
matthyx: DaemonSet profiles have the same revision semantics: `k8s-interface` sets their instance-ID name to `<daemonset-name>-<controller-revision-hash>` (`GenerateInstanceIDFromRuntimeObj`, `case "DaemonSet"`), and the profile labels only carry `workload-kind=DaemonSet` / `workload-name=<ds>`. Without a matching branch here, a stale DaemonSet-revision profile matches any present DaemonSet pod (verified with a scratch test on this branch → `true`). Suggest handling `DaemonSet` alongside `StatefulSet`: revision = `strings.TrimPrefix(instanceID.Name, name+"-")`, compared with `pod.Labels[appsv1.StatefulSetRevisionLabel]`, owner kind `DaemonSet`.

matthyx (changes requested):
Re-reviewed `c8d3e02`. Both previous blockers are resolved:

- StatefulSet revisions: parsing the `instance-id` annotation with `GenerateInstanceIDFromString` and matching `controller-revision-hash` + owner UID is correct — `k8s-interface` sets the StatefulSet instance-ID name to the pod's `controller-revision-hash` label, so the comparison lines up. The old/current/missing-revision test cases cover it.
- `podIsPresent` (only `DeletionTimestamp == nil`) restores scanning for completed/failed Job and bare-Pod profiles; the new CronJob/Job fixtures assert that.

`go build ./...`, `go vet`, `gofmt -l` clean, `go test ./config/ ./watcher/` pass locally; all CI checks are green on `c8d3e02`.

**One remaining gap of the same class — DaemonSet revisions.** `k8s-interface` handles DaemonSet pods the same way as StatefulSets (instance-ID name = `<daemonset-name>-<controller-revision-hash>`, see `GenerateInstanceIDFromRuntimeObj`), but `hasMatchingPod` only applies the revision check for `workload-kind=StatefulSet`; DaemonSet profiles fall through to "any present pod matches". A scratch test on this branch (DaemonSet with one current-revision pod, profile whose instance-id names an old revision) returns `true`, so stale DaemonSet-revision profiles are still re-dispatched on restart with the new default. Suggested fix: extend the revision branch to `DaemonSet`, comparing `pod.Labels[appsv1.StatefulSetRevisionLabel]` against `strings.TrimPrefix(instanceID.Name, name+"-")` (owner kind `DaemonSet`), plus an old/current DaemonSet pair in the relist test. Small change; happy to approve once it's in.

Reviewed commit: `c8d3e02` against `main` (`616d85b`).

ANAMASGARD:
@matthyx   Extended the revision branch to DaemonSet: TrimPrefix the workload name from the instance-id AlternateName before comparing with controller-revision-hash, plus owner UID check. Added old/current DaemonSet fixtures to the relist test.

matthyx (approved):
Reviewed `f8668ff`. The DaemonSet branch mirrors the StatefulSet one correctly: `k8s-interface` sets the DaemonSet instance-ID name to `<name>-<controller-revision-hash>`, so `strings.TrimPrefix(instanceID.Name, name+"-")` yields exactly the pod's `controller-revision-hash` label, and the owner kind/name/UID check ties it to the live workload. The old/current DaemonSet fixtures cover both outcomes.

All findings from the earlier rounds are now addressed (StatefulSet and DaemonSet revisions, completed/failed Job and Pod profiles, `matchExpressions` selectors). Locally on `f8668ff`: `go build ./...`, `go vet`, `gofmt -l` clean, `go test ./config/ ./watcher/` pass; all CI checks green.

Non-blocking follow-ups (not for this PR): `kind=Pod` profiles fetch the pod twice (`hasMatchingPod` + `getPod`); the new default is not exposed in `kubescape/helm-charts`.

Approving `f8668ff` against `main` (`616d85b`).

PR #420, opened 2026-09-21
feat(config): add regex support for namespace filtering

### LINKED ISSUE #387: Regex for includeNamespaces / excludeNamespaces
## Overview
Allow `includeNamespaces` / `excludeNamespaces` in the kubescape-operator
  config to accept regex patterns in addition to exact namespace names.

## Problem
From a scalable standpoint, it would be nice to create a regex so that we can have dynamic namespaces.

## Solution
    includeNamespaces: "kube-system"              # unchanged, exact match
    includeNamespacesRegex: "^team-.*-prod$"      # new, RE2 regex
    excludeNamespaces: [...]
    excludeNamespacesRegex: [...]

  A namespace is included if it matches any entry in either list; excluded likewise. Patterns use Go's `regexp` (RE2) and are compiled once at config load — invalid patterns fail fast with a clear error.

## Alternatives
**Flux GitOps workaround**: a CronJob with namespace list/get RBAC resolves the regex against live cluster state every N minutes and writes the matched names to a ConfigMap. The HelmRelease references it via `spec.valuesFrom`, so Flux re-renders the chart whenever the list changes. 

## Additional context
- Related: kubescape/helm-charts#664 (dynamic reconfiguration without redeploy)
- Example would be how prometheus handles namespaces for alerts

Nachiket-Roy:
## Overview

Adds RE2 regular-expression support for namespace filtering in `kubescape-operator`, resolving #387.

- **Precedence:** if either an include list or include-regex is set, inclusion mode is active — only namespaces matching an exact name or an include pattern are scanned; everything else is skipped. Otherwise, if an exclude list or exclude-regex is set, matching namespaces are skipped. If neither is set, no namespaces are skipped.
- **Fail-fast compilation:** patterns are compiled once at config load via `compileRegexes`; malformed RE2 syntax aborts operator startup with a descriptive error rather than failing silently per-event.
- **Patterns are unanchored by default** (matching anywhere in the namespace name), so `team-` matches `my-team-staging`. Use `^`/`$` to anchor.
- **RE2 constraints apply:** no lookahead, lookbehind, or backreferences.
- **Comma handling:** exact-match lists are comma-separated; when a regex itself contains a comma, use a JSON array / YAML list.

## Additional Information

- Regexes are compiled into `[]*regexp.Regexp` on `OperatorConfig` at initialization so the hot path (`SkipNamespace`, called per event) only runs `MatchString` — no per-call compilation.
- `compileRegexes` trims whitespace and skips empty strings, so config defaults of `""` are inert.
- Introspection getters (`IncludeNamespacesRegex()`, `ExcludeNamespacesRegex()`, `IncludeNamespaces()`, `ExcludeNamespaces()`) expose raw slices for logging and test assertions.
- `configuration/config.json` gains default empty `excludeNamespacesRegex` / `includeNamespacesRegex` fields for parity with `includeNamespaces`.

## How to Test

1. **Config package tests:**
   ```bash
   go test -v ./config/...
   ```
   Covers `TestLoadConfig_Regex` (comma string, JSON array, fail-fast on bad syntax), `TestOperatorConfig_SkipNamespace` (21 scenarios), and `TestOperatorConfig_Getters`.

2. **Vet / lint:**
   ```bash
   go vet ./config/...
   ```

3. **Manual check (optional):** set `includeNamespacesRegex: ["^team-"]` in the operator config, deploy, and confirm only namespaces beginning with `team-` are scanned. Then set `excludeNamespacesRegex: ["-kube-system$"]` and confirm matching namespaces are skipped.

## Related issues/PRs

- Resolved #387

- [x] If this is a core feature, I have added thorough tests.

## Summary by CodeRabbit

* **New Features**
  * Added exact-match and regular-expression filters for including or excluding namespaces.
  * Include rules take precedence over exclusion rules.
  * Invalid regular-expression patterns are reported during configuration loading.
  * Added default configuration examples for the new namespace filters.

* **Documentation**
  * Documented filtering precedence, regex behavior, supported limitations, and JSON array formatting requirements.


README.md
```diff
@@ -304,6 +308,24 @@ The Operator reads configuration from `/etc/config/`. When running locally, set
...
+| `includeNamespacesRegex` | `""` | Comma-separated string or array of RE2 regex patterns for namespaces to include |
+| `excludeNamespaces` | `"kube-system,kubescape"` | Comma-separated string or array of exact namespaces to exclude |
+| `excludeNamespacesRegex` | `""` | Comma-separated string or array of RE2 regex patterns for namespaces to exclude |
+
+### Namespace Filtering (Exact & Regex)
+
+The Operator supports filtering namespaces using exact names (`includeNamespaces`, `excludeNamespaces`) and RE2 regular expressions (`includeNamespacesRegex`, `excludeNamespacesRegex`):
+
+- **Precedence (Allow-List)**: If either `includeNamespaces` or `includeNamespacesRegex` is configured, inclusion mode is active. Only namespaces matching an exact entry in `includeNamespaces` OR a regex in `includeNamespacesRegex` will be processed. All other namespaces are skipped, and `excludeNamespaces` / `excludeNamespacesRegex` are ignored.
+- **Exclusion (Deny-List)**: When no inclusion rules are set, any namespace matching an exact entry in `excludeNamespaces` OR a regex in `excludeNamespacesRegex` will be skipped.
```
matthyx, line 321: [HIGH] The documented guarantee is not true for ContainerProfile scans. `HandleContainerProfileEvents` (`watcher/containerprofilewatcher.go:104-157`) sends every eligible profile without calling `SkipNamespace`; its watcher explicitly relies on node-agent filtering, but the pinned node-agent config supports only the exact include/exclude lists. With `EXCLUDENAMESPACESREGEX=^systest-ns-rarz$` loaded, the existing ContainerProfile test still emitted three scan commands for `systest-ns-rarz`. This can scan namespaces users explicitly excluded. Apply `wh.cfg.SkipNamespace(obj.Namespace)` before pod lookup/command creation and add excluded-regex, outside-include-regex, and allowed control cases.

config/config.go
```diff
@@ -182,22 +185,44 @@ type IConfig interface {
...
+			return nil, fmt.Errorf("invalid regex pattern %q: %w", p, err)
+		}
+		compiled = append(compiled, re)
+	}
+	return compiled, nil
+}
+
 func NewOperatorConfig(components CapabilitiesConfig, clusterConfig utilsmetadata.ClusterConfig, creds *utils.Credentials, serviceConfig Config) *OperatorConfig {
+	incRegex, _ := compileRegexes(serviceConfig.IncludeNamespacesRegex)
+	excRegex, _ := compileRegexes(serviceConfig.ExcludeNamespacesRegex)
```
matthyx: [MEDIUM] Please do not discard these compilation errors. `compileRegexes` returns `nil` as soon as any pattern is invalid, so a direct `Config{IncludeNamespacesRegex: []string{"^prod$", "["}}` silently loses the valid restriction and `SkipNamespace("dev")` allows the namespace. `main` is currently protected by `LoadConfig`, but this exported constructor fails open and contradicts the fail-fast contract. Propagate the error (or centralize validation/compilation in one error-returning construction path) and cover mixed valid/invalid include and exclude inputs.

matthyx (changes requested):
Requesting changes on `36889611e66352b256e58e39db5ea4eb37b27588` (base `main` at `58cffc9ed12359205232bde43477adfe740bf612`).

The feature is needed: #387 describes a real scalability gap, and a maintainer supported the general direction. The core predicate is efficient and its exact/regex union, include precedence, RE2 behavior, and load-time validation are well tested. However, two correctness gaps remain:

- **HIGH — regex exclusions are bypassed by ContainerProfile scans.** The new documentation promises namespace-wide filtering, but `HandleContainerProfileEvents` dispatches scans without `SkipNamespace`, relying on node-agent filtering that supports only the existing exact lists. In an isolated run with `EXCLUDENAMESPACESREGEX=^systest-ns-rarz$`, the configuration loader received that regex and the ContainerProfile test still emitted three scan commands for `systest-ns-rarz`. This makes the advertised policy inconsistent across scan paths.
- **MEDIUM — the exported constructor silently drops invalid regex policy.** `NewOperatorConfig` ignores both compilation errors. A directly supplied mixed valid/invalid list becomes `nil`, which can fail open. The current `main` path is protected by `LoadConfig`, but the constructor itself does not honor the fail-fast contract.

History/search: merged #230 introduced exact include/exclude filtering; merged #405 fixes the separate `continuousScanning.matchingRules.namespaces` path. Neither implements regexes or supersedes this PR. #387 contains an unreviewed proposal to resolve regexes in Helm; it was neither explicitly accepted nor rejected. Searches covered open/closed PRs and issues in this repository using `includeNamespacesRegex`, `includeNamespaces`, `excludeNamespaces`, `SkipNamespace`, `namespace regex`, and `namespace filtering`, plus git history; GitHub indexing/search limits mean no further match is not proof none exists.

Validation:

- `go test -v ./config/...` — passed in a disposable, read-only-source container.
- `go vet ./config/...` — passed in the same isolation model.
- Targeted ContainerProfile reproduction with the exclusion environment override — passed its existing assertions while logging scans for the excluded namespace, confirming the gap.
- Whole `./watcher` package — not completed: its k3s test requires a Docker daemon unavailable inside the hardened container; the focused watcher test did run.
- `git diff --check` / `gofmt -d` found only nonblocking test-file formatting (alignment and a trailing blank line).
- Current GitHub checks cover DCO, dependency/license quality, GitGuardian, and CodeRabbit; no Go unit-test check is reported for this head.

The prior CodeRabbit README-default comment is resolved at this head and is not repeated. Please add the namespace guard plus regression coverage, and make regex compilation errors impossible to discard before re-review.

matthyx (approved):
Approved at `d7551a78342ccfb525e48b8cd17c5d83a15a6326` against `main` `58cffc9ed12359205232bde43477adfe740bf612`.

Both prior blockers are fixed:

- ContainerProfile events now apply `SkipNamespace` before any pod lookup or command dispatch, with excluded, nonmatching-include, and allowed controls covered.
- `NewOperatorConfig` now fails closed with a field-specific error, and every repository caller handles the new error result. Mixed valid/invalid regex tests confirm no partial policy escapes.

Validation performed in a disposable container with read-only source and no host credentials: `go test -v ./config/...`, `go test -run '^TestHandleContainerProfileEvents$' ./watcher`, and `go test ./mainhandler ./continuousscanning` all passed; `go vet ./config/... ./watcher ./mainhandler ./continuousscanning` passed. Diff whitespace and Go formatting checks are clean. The isolated full watcher-package suite remains outside this evidence because it requires nested Docker for its k3s test.

The current CodeRabbit slice-aliasing note is a nonblocking API-hygiene suggestion; it has no demonstrated production caller or policy bypass. No material blockers remain.

